PR #3, opened 2021-12-28
[WIP] Connection pooling with deadpool

vimmerru:
Here is PoC for connection pooling based on deadpool create. It mostly mirrors `r2d2` code from main diesel repo. If you have an interest to contribution i can finalise PR and provide tests.

The code in theory can be used outside of the repo, but it will require wrapping of connections to be able implement traits and there is enough traits magic to make it not so simple task.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -340,6 +340,24 @@ dependencies = [
 ]
 
 [[package]]
+name = "deadpool"
+version = "0.9.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3bf0c5365c0925c80a838a6810a1bf38d3304ca6b4eb25829e29e33da12de786"
+dependencies = [
+ "async-trait",
+ "deadpool-runtime",
+ "num_cpus",
+ "tokio",
+]
+
+[[package]]
+name = "deadpool-runtime"
+version = "0.1.2"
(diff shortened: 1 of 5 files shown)


weiznich:
Thanks for opening this PR :+1: I will try to have a look at this in the first weeks of the new year.

joshbenaron:
@vimmerru Could you also include an example usage please?

vimmerru:
@joshbenaron I added usage example in tests/deadpool.rs. It uses similar test code that your original test, but for now is dedicated as i am not sure what is right way to integrate with your existing CI.

joshbenaron:
```the trait `ManagedAsyncConnection` is not implemented for `diesel_async::deadpool::ConnectionManager<AsyncPgConnection>```

I'm getting this error. I might be being stupid

vimmerru:
@joshbenaron Do you have it when compile tests? ManagedAsyncConnection should implemented for managed objects, not for manager itself.

vimmerru:
@joshbenaron Can you also provide me exact command you are using? Maybe there is some issue with features combination.

weiznich:
I just want to leave a note here that I've not forgot about this PR. Unfortunately I do currently not have the time to look into this in detail, hopefully that will become better in the next weeks.

PR #14, opened 2022-07-23
Fix a clippy warning

Razican:
This PR fixes a small clippy warning that was appearing in the CI. Basically, since the result is a unit structure, it doesn't make sense to bind it, even if it's to a discarded binding.


weiznich:
Fixed as part of #22

PR #19, opened 2022-09-01
Using Diesel 2.0.0

Razican:
This PR updates diesel to use the recently released 2.0 stable version, instead of the RC1. It also removes the Cargo.lock file from the repository, so it overrides #16.


weiznich:
Thanks for opening this PR. I'm on vacation for the next weeks and will have a look at this after I'm back.

weiznich:
Fixed as part of #22

Razican:
> Fixed as part of #22

Note that the dependencies now use Diesel 2.0.0, but the dev-dependencies still use the RC.

PR #24, opened 2022-09-28
Update dependencies / Cargo.toml

Bajix:
- Tokio updated to use "1" for compatibility + make rt-multi-thread dev-dependency
- Diesel dev-dependency set as 2.0.0 (erroneously 2.0.0-RC-1)
- Remove tokio/rt-multi-thread from postgres feature (needed only for testing)
- Alpha sort dependencies

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -45,9 +45,9 @@ checksum = "8da52d66c7071e2e3fa2a1e5c6d088fec47b593032b254f5e980de8ea54454d6"
 
 [[package]]
 name = "async-trait"
-version = "0.1.56"
+version = "0.1.57"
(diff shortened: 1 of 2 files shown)


weiznich:
Essentially fixed by an improved version of this in #29

Bajix:
I think the optional tokio dependency still needs the rt feature

weiznich:
@Bajix I just rechecked that everything compiles with the current feature send. I didn't find any issue. What's the reason why you believe the `rt` feature is required?

Bajix:
tokio::task::spawn is used in pg/mod.rs and so the postgres feature flag needs to enable tokio with rt otherwise spawn can’t be used in non-dev

PR #28, opened 2022-10-04
mysql: close last_stmt for non-cacheable statements

### LINKED ISSUE #26: MySQL: `Can't create more than max_prepared_stmt_count statements` after executing more than 16382 inserts
diesel-async + AsyncMysqlConnection gives error after executing more than 16382 inserts, while diesel does not
```
Can't create more than max_prepared_stmt_count statements (current value: 16382)
```

<details>
<summary> outdated reprodicible example </summary>
Hi. I'm setting up a really simple [template repo](https://github.com/gwy15/diesel-async-example) for my own use and I tried to [insert-select-delete](https://github.com/gwy15/diesel-async-example/blob/main/tests/insert_and_validate.rs#L51) concurrently.

<del>Specifically, I used 100 threads/futures, each insert-select-delete 200 times. The thread pool was set max connection count to 30.</del>

diesel-async gives the error, whilst disel(sync) does not and can finish as expected.
```
Can't create more than max_prepared_stmt_count statements (current value: 16382)
```

<del>On my local machine they both reached about ~2k QPS. The concurrent connections was 30.</del>

reproducible repo (with corresponding failing CI): https://github.com/gwy15/diesel-async-example

</details>

### Versions

- **Rust:** stable rustc 1.64.0 (a55dd71d5 2022-09-19)
- **Diesel:** 2
- **Diesel_async:** 0.1.0
- **Database:** mysql
- **Operating System** windows/linux

### Feature Flags

- **diesel:** ["mysql", "chrono"]
- **diesel_async:** ["bb8", "mysql", "tokio"]

Since diesel + multithread does not have a problem, I expect diesel-async to have the same ability.

`Can't create more than max_prepared_stmt_count statements`

No

### Steps to reproduce
`migration`
```
CREATE TABLE `user` (
    `id`    bigint unsigned     NOT NULL AUTO_INCREMENT,
    `name`  varchar(255)        NOT NULL,
    `email` varchar(255)        NOT NULL,
    `created_at` datetime(3) NOT NULL DEFAULT now(3),
    `updated_at` datetime(3) NOT NULL DEFAULT now(3) ON UPDATE now(3),
    PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
```

`main.rs`
```
use diesel_async::{AsyncConnection, AsyncMysqlConnection, RunQueryDsl};

mod schema {
    diesel::table! {
        user (id) {
            id -> Unsigned<Bigint>,
            name -> Varchar,
            email -> Varchar,
            created_at -> Datetime,
            updated_at -> Datetime,
        }
    }
}

#[derive(diesel::Insertable)]
#[diesel(table_name = schema::user)]
pub struct NewUser<'a> {
    pub name: &'a str,
    pub email: &'a str,
}
impl<'a> NewUser<'a> {
    pub async fn insert(&self, conn: &mut AsyncMysqlConnection) -> anyhow::Result<()> {
        diesel::insert_into(schema::user::dsl::user)
            .values(self)
            .execute(conn)
            .await?;
        Ok(())
    }
}

#[tokio::main]
async fn main() -> anyhow::Result<()> {
    let url = dotenv::var("DATABASE_URL")?;

    let mut conn = diesel_async::AsyncMysqlConnection::establish(&url).await?;

    for i in 0..16382 + 10 {
        NewUser {
            name: &format!("name-{}", i),
            email: &format!("email-{}", i),
        }
        .insert(&mut conn)
        .await?;
        if i % 100 == 0 {
            println!("{}", i);
        }
    }
    Ok(())
}
```

`cargo.toml`
```
[dependencies]
anyhow = "1.0.65"
diesel = { version = "2.0.0", features = ["mysql", "chrono"] }
diesel-async = { version = "0.1.0", features = ["bb8", "mysql", "tokio"], optional = true }
dotenv = "0.15.0"
tokio = { version = "1.21.2", features = ["full"] }
```

<details>
<summary>outdated</summary>
Please see my repo https://github.com/gwy15/diesel-async-example, and the CI workflow https://github.com/gwy15/diesel-async-example/actions/runs/3175713190 is exactly what I had on my local machine.

to run locally, please
```bash
git clone https://github.com/gwy15/diesel-async-example
cd diesel-async-example
docker-compose up -d
diesel migration run
cargo t --no-default-features --features async
```
</details>

## Checklist

gwy15:
Since we disabled mysql_async's caching mechanism and in that way [it requires manual close the statement](https://docs.rs/mysql_async/latest/mysql_async/struct.Opts.html#method.stmt_cache_size), we need to close all prepared stmts before dropping it.

From my understanding, and correct me if I'm wrong, as long as statements are cacheable, we will never drop them. So we only need to drop non-cacheable stmts. This PR implements this.

## Concern
Is it better to close prepare statements right after execution or close before the next execution of non-cacheable statement (the implementation)?

close #26

### DIFF
--- a/src/mysql/mod.rs
+++ b/src/mysql/mod.rs
@@ -22,11 +22,15 @@ use self::row::MysqlRow;
 use self::serialize::ToSqlHelper;
 
 /// A connection to a MySQL database. Connection URLs should be in the form
-/// `mysql://[user[:password]@]host/database_name`
+/// `mysql://[user[:password]@]host/database_name`.
+///
+/// See [`mysql_async::Opts`] for more available connection url queries. Notice that
+/// param `stmt_cache_size` will be ignored.
 pub struct AsyncMysqlConnection {
     conn: mysql_async::Conn,
     stmt_cache: StmtCache<Mysql, Statement>,
     transaction_manager: AnsiTransactionManager,
+    // a workaround so that for non-cacheable stmts we still get a `&'conn Statement`.
     last_stmt: Option<Statement>,
 }
 
@@ -212,11 +216,14 @@ impl AsyncMysqlConnection {
 
         let stmt = stmt_cache.cached_prepared_statement(query, &metadata, conn, &Mysql);
 
-        stmt.and_then(|(stmt, conn)|async  move {
-
+        stmt.and_then(|(stmt, conn)| async move {
             let stmt = match stmt {
                 MaybeCached::CannotCache(stmt) => {
+                    if let Some(last_stmt) = std::mem::take(last_stmt) {
+                        conn.close(last_stmt).await.map_err(ErrorHelper)?;
+                    }
                     *last_stmt = Some(stmt);
+                    // SAFETY: we just set last_stmt to Some.
                     last_stmt.as_ref().unwrap()
                 }
                 MaybeCached::Cached(s) => s,


gwy15:
diesel seems to guarantee that [For queries which are unsafe to cache, ... will be prepared and immediately released after use](https://github.com/diesel-rs/diesel/blob/master/diesel/src/connection/statement_cache.rs#L46)

In that way I think `last_stmt` should be dropped and `Queryable::close`-ed right after executing `callback`. But the returned `R` has lifetime `'conn` and Rust think it borrowed `conn` and `last_stmt` which makes it impossible to use `conn` and `last_stmt` again...

weiznich:
Thanks for opening this PR. I've filled #29 with a more complete solution. I will likely merge the other PR.

weiznich:
Closed by #29

PR #30, opened 2022-10-07
Fix the type alias for deadpool's `BuildError`

fluxxu:
If I understand it correctly, `diesel_async::pooled_connection::deadpool::BuildError` is the error returned when
`deadpool::managed::PoolBuilder::build` failed. 
According to the [docs](https://docs.rs/deadpool/0.9.5/deadpool/managed/struct.PoolBuilder.html), the error should be `BuildError<M::Error>` where M is `AsyncDieselConnectionManager`, so the error should be
`deadpool::managed::BuildError<<AsyncDieselConnectionManager as deadpool::managed::Manager>::Error>` which is 

`deadpool::managed::BuildError<diesel_async::pooled_connection::PoolError>`, 

but in the current code, it is an alias of 

`deadpool::managed::BuildError<deadpool::managed::PoolError<diesel_async::pooled_connection::PoolError>>`

### DIFF
--- a/src/pooled_connection/deadpool.rs
+++ b/src/pooled_connection/deadpool.rs
@@ -47,7 +47,7 @@ pub type Pool<C> = deadpool::managed::Pool<AsyncDieselConnectionManager<C>>;
 /// Type alias for using [`deadpool::managed::PoolBuilder`] with [`diesel-async`]
 pub type PoolBuilder<C> = deadpool::managed::PoolBuilder<AsyncDieselConnectionManager<C>>;
 /// Type alias for using [`deadpool::managed::BuildError`] with [`diesel-async`]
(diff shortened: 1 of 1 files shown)


weiznich:
Thanks for opening this PR. I potentially plan to relicense diesel-async in the future. Because of that I need an explicitly written consent that you transfer the ownership of this change to me.

weiznich:
Thanks for opening this PR. I potentially plan to relicense diesel-async in the future. Because of that I need an explicitly written consent that you transfer the ownership of this change to me. Please add that to the commit message.

(Hit the wrong button…)

fluxxu:
I consent that I transfer the ownership of this change to you.

PR #31, opened 2022-10-12
Removed lock file from Git

Razican:
This PR removes the `Cargo.lock` file from Git. It was already in the `.gitignore` file, but it was still being tracked.


weiznich:
Fixed by #55

PR #36, opened 2022-10-26
Add rudimentary TLS support for postgresql

johnchildren:
Adds a postgresql-native-tls dependency and uses it by default. This brings the postgresql inline with mysql_async which also depends on native_tls.

---

I am not sure if this is the best way to implement this feature, I can re-work it to be behind a feature flag if required.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -357,7 +357,7 @@ dependencies = [
 
 [[package]]
 name = "diesel-async"
-version = "0.1.0"
+version = "0.1.1"
 dependencies = [
  "async-trait",
  "bb8",
@@ -369,6 +369,8 @@ dependencies = [
  "mobc",
  "mysql_async",
  "mysql_common",
+ "native-tls",
+ "postgres-native-tls",
  "tokio",
  "tokio-postgres",
 ]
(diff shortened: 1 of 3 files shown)


weiznich:
Thanks for opening this PR. I'm not interested in adding the tls logic to the `establish` method. You can construct a new `AsyncPgConnection` form an existing `tokio_postgres::Client` via [`AsyncPgConnection::try_from`](https://docs.rs/diesel-async/0.1.1/diesel_async/struct.AsyncPgConnection.html#method.try_from) method. This allows to customize the setup as required.

johnchildren:
> Thanks for opening this PR. I'm not interested in adding the tls logic to the `establish` method. You can construct a new `AsyncPgConnection` form an existing `tokio_postgres::Client` via [`AsyncPgConnection::try_from`](https://docs.rs/diesel-async/0.1.1/diesel_async/struct.AsyncPgConnection.html#method.try_from) method. This allows to customize the setup as required.

That makes sense, thank you, though I am a little unclear on how this approach works with the connection pooling feature which seems to rely on `establish`. Would the recommended approach there be to re-implement the connection managers to use the customized setup?

weiznich:
As far as there is something like a "recommended approach" for that yet, that's re-implementing the connection manager. I might address that in a future release by providing a more general connection manager implementation, but that likely takes some time as I currently have other priorities.

PR #38, opened 2022-11-08
add usage of scoped futures to transactions: ownership of this change transferred to @weiznich

treysidechain:
Hi there! I've been building on top of the `diesel-async` library for some time and have wanted to be able to use references in the closures I pass to `AsyncConnection`'s transaction method. I wrote up a small [crate](https://crates.io/crates/scoped-futures) that encapsulates the behavior needed to support references in the closure, basically it imposes an upper bound on the higher-ranked lifetime that's implicit in the FnOnce trait bound so that it doesn't imply a `'static` bound.

Main downside of merging this would be pushing a new function `scope_boxed` onto code calling `transaction`, however, `Box::pinned` still works with the proposed changes.

Here is an [example](https://docs.rs/scoped-futures/latest/scoped_futures/trait.ScopedFuture.html#example) showing how the updated `transaction` signature allows callback closures to capture references.


weiznich:
Thanks for opening this PR, this looks great :+1:
I would like to reexport the `ScopedFutureExt` trait from `diesel_async` so that it is easier for users to import this trait.

In addition: I potentially plan to relicense diesel-async in the future. Because of that I need an explicitly written consent that you transfer the ownership of this change to me. Please add that to the commit message.

treysidechain:
sure thing, I updated the PR title. If that's not what you're looking for let me know

Issue #39, opened 2022-11-16
Relicense the crate under MIT/Apache 2.0


weiznich:
This issue tracks the progress to relicense `diesel_async` from AGPL 3.0 to MIT/Apache 2.0.

I merged contributions from the following people:

- [x] @smklein 
    +  https://github.com/weiznich/diesel_async/pull/8 
- [x] @Razican 
    + https://github.com/weiznich/diesel_async/pull/15
- [ ] @gwy15
    + https://github.com/weiznich/diesel_async/pull/21
- [x] @fluxxu 
    + https://github.com/weiznich/diesel_async/pull/30 (already consent to transfer the ownership of this change to me)
- [x] @johnchildren 
    + https://github.com/weiznich/diesel_async/pull/35 (already consent to transfer the ownership of this change to me)
- [x] @livingcorpse25 
    + https://github.com/weiznich/diesel_async/pull/37 (already consent to transfer the ownership of this change to me)
- [x] @treysidechain 
    + https://github.com/weiznich/diesel_async/pull/38 (already consent to transfer the ownership of this change to me)
- [x] @medcat
    + https://github.com/weiznich/diesel_async/pull/40 (already consent to transfer the ownership of this change to me)

For @smklein, @Razican  and @gwy15: Please comment on this issue whether you agree to relicense your code or not?
For @fluxxu, @johnchildren, @livingcorpse25, @treysidechain: I already checked your boxes above, as you agreed to transfer the ownership of your changes to me. Please comment if you disagree with this change, otherwise I assume your consent.

smklein:
I agree to relicense (and I'll check my own box). Thank you for doing this!

Razican:
I also agree to relicense the code :)

PR #40, opened 2022-11-25
Re-export of TransactionManagerStatus

medcat:
Not sure if this is intended, but `TransactionManagerStatus` is effectively hidden from public access; however, this enum is required if you want to implement the trait `TransactionManager`, as the trait function `TransactionManager::transaction_manager_status_mut` returns a `TransactionManagerStatus` directly.  As far as I can tell, it is accessible in [diesel], even if gated behind a feature.

[diesel]: https://github.com/diesel-rs/diesel/blob/master/diesel/src/connection/mod.rs#L18

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -95,7 +95,9 @@ pub use self::pg::AsyncPgConnection;
 pub use self::run_query_dsl::*;
 
 #[doc(inline)]
-pub use self::transaction_manager::{AnsiTransactionManager, TransactionManager};
+pub use self::transaction_manager::{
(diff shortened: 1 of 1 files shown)


weiznich:
Thanks for opening this PR :+1:

I potentially plan to relicense diesel-async in the future. (See #39)  Because of that I need an explicitly written consent that you transfer the ownership of this change to me. Please add that to the commit message.

weiznich (approved):

weiznich:
Yes that should be fine :+1:

PR #46, opened 2022-12-16
Prefer sub-crates of `futures`

nickelc:
This removes `futures-executor` and `futures-io` from the transient dependencies.


Cargo.toml
```diff
@@ -15,7 +15,8 @@ description = "An async extension for Diesel the safe, extensible ORM and Query
 [dependencies]
 diesel = { version = "2.0.0", default-features = false, features = ["i-implement-a-third-party-backend-and-opt-into-breaking-changes"]}
 async-trait = "0.1.51"
-futures = "0.3.17"
+futures-channel = { version = "0.3.17", default-features = false, features = ["std", "sink"] }
```
weiznich, line 18: Channels are currently only used by the mysql connection implementation. It would be meaningful to mark this dependency as optional and put it behind the `mysql` feature flag.

weiznich (changes requested):
Thanks for submitting this PR. If we work on reducing the dependency count we could go a step further and put the `futures-channel` dependency behind the `mysql` feature flag as this crate is only used there.

weiznich (approved):
Thanks :+1:

PR #48, opened 2022-12-18
Add fmt::Debug impls to AsyncDieselConnectionManager and AsyncPgConnection

asaaki:
First, thank you for this crate and also the latest changes (`new_with_setup`), which enabled me to use it with rustls.

-----

This change is needed to use the `#[instrument]` macro of [tracing](https://crates.io/crates/tracing), which seems to require Debug impls for types being part of the function; since they get printed as part of the logs.

Example:
https://github.com/asaaki/capture-the-ip/blob/main/cti_core/src/app/api.rs#L95-L97

The state is a pool represented by the above mentioned types.
The tracing crate is very unhappy if those types are not Debug.

I kept the implementations very simple and minimal;
the connection would only display its name, the manager also includes the connection URL. All other fields left untouched to not infect everything.

### DIFF
--- a/src/pg/mod.rs
+++ b/src/pg/mod.rs
@@ -103,7 +103,7 @@ pub struct AsyncPgConnection {
 
 impl std::fmt::Debug for AsyncPgConnection {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        f.debug_struct("AsyncPgConnection").finish()
+        f.debug_struct("AsyncPgConnection").finish_non_exhaustive()
     }
 }
 
--- a/src/pg/mod.rs
+++ b/src/pg/mod.rs
@@ -101,6 +101,12 @@ pub struct AsyncPgConnection {
     metadata_cache: Arc<Mutex<Option<PgMetadataCache>>>,
 }
 
+impl std::fmt::Debug for AsyncPgConnection {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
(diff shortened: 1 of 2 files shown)


kaj:
Having debug impls on those seems nice to me, so I second this.  But the connection URL sometimes (often?) contains a password, so maybe including that is risky?

asaaki:
I'm also okay with removing the connection string from it or replace it with `"redacted"`.
At least for me I don't need these instances to print anything meaningful during request tracing/debugging.

weiznich:
Thanks for opening this PR :+1:

I would prefer to keep that similar to the connection(pool) impls in diesel. These do also not offer debug impls, especially as there is nothing really helpful what could be printed in such debug impl anyway.

> This change is needed to use the #[instrument] macro of [tracing](https://crates.io/crates/tracing), which seems to require Debug impls for types being part of the function; since they get printed as part of the logs.

You can easily ignore parameters without debug impl by using [`#[instrument(skip(your_ignored_param)]`](https://docs.rs/tracing-attributes/latest/tracing_attributes/attr.instrument.html#skipping-fields)

asaaki:
> You can easily ignore parameters without debug impl by using [#[instrument(skip(your_ignored_param)]](https://docs.rs/tracing-attributes/latest/tracing_attributes/attr.instrument.html#skipping-fields)

OMG, thanks for pointing that out. Should read the docs more thoroughly next time. 😅 

Adjusted my code accordingly and it seems to work without the Debug hack.

Sorry for the noise; closing the PR. 😬

PR #64, opened 2023-03-10
Add an example of using a pooled bb8 connection with rustls

ThouCheese:


examples/postgres/pooled-with-rustls/src/main.rs
```diff
@@ -0,0 +1,62 @@
...
+    Ok(())
+}
+
+fn establish_connection(config: &str) -> BoxFuture<ConnectionResult<AsyncPgConnection>> {
+    let fut = async {
+        // We first set up the way we want rustls to work.
+        let rustls_config = rustls::ClientConfig::builder()
+            .with_safe_defaults()
+            .with_root_certificates(root_certs())
+            .with_no_client_auth();
```
weiznich, line 41: I think it might be a good idea to add a comment here that this is the equivalent setup for libpq's `sslmode=verify-full` so that the next person has an easier time to figure out why this not works with their certs, while it works with `libpq`.

examples/postgres/pooled-with-rustls/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "pooled-with-rustls"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+diesel = { version = "2.0.3", default-features = false, features = ["postgres"] }
+diesel-async = { version = "0.2.0", features = ["bb8", "postgres"] }
```
weiznich, line 10: So that it picks up the diesel-async version from this repository

```suggestion
diesel-async = { path= "../../",  version = "0.2.0", features = ["bb8", "postgres"] }
```

weiznich (commented):
Thanks for adding this example :+1:

Beside the minor change to the example I would like integrate the example into repository. That means:

* Add it to the workspace here: https://github.com/weiznich/diesel_async/blob/8d0857503d568d238b9e73d2a22bb652c08e796c/Cargo.toml (requires introducing a workspace item there first)
* Add it to the CI setup here: https://github.com/weiznich/diesel_async/blob/8d0857503d568d238b9e73d2a22bb652c08e796c/.github/workflows/ci.yml#L117

weiznich (approved):

PR #68, opened 2023-03-31
Fix deadpool example in README

dmeijboom:
While setting up diesel async in axum I noticed that the README had a tiny error. Looking at the docs of deadpool the `Pool::builder(..)` accepts the `manager` as the first argument (instead of the `build()` method).


weiznich (approved):
Thanks :+1:

PR #92, opened 2023-07-16
fix mysql `TinyInt` serialization

### LINKED ISSUE #91: Incorrect casting of MySQL `tinyint`
### Versions

- **Rust:** rustc 1.70.0 (90c541806 2023-05-31)
- **Diesel:** 2.1.0
- **Diesel_async:** 0.3.1
- **Database:** mysql 8.0

### Feature Flags

- **diesel:** `["mysql_backend", "serde_json", "uuid"]`
- **diesel_async:** `["mysql", "mobc"]`

## Problem Description
having such a table:
```rust
diesel::table! {
    use diesel::sql_types::*;
    use super::sql_types::StateMapping;

    accounts (address, workchain) {
        workchain -> Tinyint,
        #[max_length = 32]
        address -> Varbinary,
        ...
    }
}
```

this query
```rs
#[derive(Queryable, Selectable)]
#[diesel(check_for_backend(Mysql))]
#[diesel(table_name = crate::schema::accounts)]
pub struct AccountWithCreatorInfo {
   ...
}

 let query = accounts_dsl::accounts
            .filter(accounts_dsl::workchain.eq(-1))
            .select(AccountWithCreatorInfo::as_select());
```

produces query like this:
```rs
"SELECT  ... FROM `accounts` WHERE (`accounts`.`workchain` = ?)",
    binds: [
        -1,
    ],
```

But [here](https://github.com/weiznich/diesel_async/blob/main/src/mysql/serialize.rs#L15) we receive values as `Vec<u8>`,
take the first element and cast them as `i64`. Casting `u8::MAX as i64` returns `255` instead of original `-1` bind.

I suggest changing this line to
```rs
            MysqlType::Tiny => Value::Int((bind[0] as i8) as i64),
```

## Checklist

0xdeafbeef:
fixes #91

### DIFF
--- a/src/mysql/serialize.rs
+++ b/src/mysql/serialize.rs
@@ -12,7 +12,7 @@ pub(super) struct ToSqlHelper {
 fn to_value((metadata, bind): (MysqlType, Option<Vec<u8>>)) -> Value {
     match bind {
         Some(bind) => match metadata {
-            MysqlType::Tiny => Value::Int(bind[0] as _),
+            MysqlType::Tiny => Value::Int((bind[0] as i8) as i64),
             MysqlType::Short => Value::Int(i16::from_ne_bytes(bind.try_into().unwrap()) as _),
             MysqlType::Long => Value::Int(i32::from_ne_bytes(bind.try_into().unwrap()) as _),
             MysqlType::LongLong => Value::Int(i64::from_ne_bytes(bind.try_into().unwrap())),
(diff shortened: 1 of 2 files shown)


weiznich (commented):
Thanks for writing this fix and including a test. I would like to simplify the test in such a way that it does not require the additional table setup.

0xdeafbeef:
> Thanks for writing this fix and including a test. I would like to simplify the test in such a way that it does not require the additional table setup.

Fixed :)

weiznich (approved):

PR #97, opened 2023-07-21
Use is_broken_transaction_manager when returning pooled item

### LINKED ISSUE #96: Dangling transactions when pooled connections dropped
### Versions

- **Diesel_async:** 0.3.1

### Feature Flags

- **diesel_async:** deadpool, postgres

## Problem Description

Firstly, thank you for your work on this crate and on Diesel itself! They have been very nice to work with compared to using tokio_postgres directly.

I gather you're aware of this issue already (https://www.reddit.com/r/rust/comments/zn9ut0/announcing_dieselasync_020/), but as dangling transactions can cause deadlocks and data loss, I'm wondering whether it might be possible to improve the current behaviour, or at least document it more prominently if that is not possible.

As a quick-and-dirty solution, would something like the following make sense, where the connection is dropped from the pool (with implicit rollback) when it is recycled with an active transaction?

```diff
diff --git a/src/pg/mod.rs b/src/pg/mod.rs
index 6a4832b..b16a0a0 100644
--- a/src/pg/mod.rs
+++ b/src/pg/mod.rs
@@ -475,6 +475,15 @@ async fn lookup_type(
 
 #[cfg(any(feature = "deadpool", feature = "bb8", feature = "mobc"))]
 impl crate::pooled_connection::PoolableConnection for AsyncPgConnection {}
+impl crate::pooled_connection::PoolableConnection for AsyncPgConnection {
+    fn is_broken(&mut self) -> bool {
+        match self.transaction_state().status.transaction_depth() {
+            Ok(Some(_)) => true,
+            Ok(None) => false,
+            Err(_) => true,
+        }
+    }
+}
 
 #[cfg(test)]
 pub mod tests {
diff --git a/src/pooled_connection/mod.rs b/src/pooled_connection/mod.rs
index ae5abdf..9c35d06 100644
--- a/src/pooled_connection/mod.rs
+++ b/src/pooled_connection/mod.rs
@@ -258,7 +258,7 @@ pub trait PoolableConnection: AsyncConnection {
     /// [ManageConnection::has_broken] for details.
     ///
     /// The default implementation does not consider any connection as broken
-    fn is_broken(&self) -> bool {
+    fn is_broken(&mut self) -> bool {
         false
     }
 }
```

Alternatively, could an approach like tokio_postgres uses work instead?
https://github.com/sfackler/rust-postgres/pull/835/files?

### Steps to reproduce

Anything that causes a transaction future to be dropped before being polled to completion, eg the calling future terminating due to client disconnection, or things like timeouts

```rust
            let mut pg = pool.get().await?;
            tokio::time::timeout(std::time::Duration::from_secs(1),
                pg.transaction::<_, Error, _>(|pg| async move {
                    sql_query("some sql").execute(pg).await?;
                    tokio::time::sleep(std::time::Duration::from_secs(2)).await;
                    Ok(())
                }.scope_boxed())
            ).await.unwrap_err();
```

## Checklist

- [x] This issue can be reproduced without requiring a third party crate (except deadpool)

dae:
Re: https://github.com/weiznich/diesel_async/issues/96#issuecomment-1645062462, is this what you had in mind?

Tentatively closes #96

### DIFF
--- a/src/pooled_connection/mod.rs
+++ b/src/pooled_connection/mod.rs
@@ -257,8 +257,9 @@ pub trait PoolableConnection: AsyncConnection {
     /// if the connection is considered to be broken or not. See
     /// [ManageConnection::has_broken] for details.
     ///
-    /// The default implementation does not consider any connection as broken
-    fn is_broken(&self) -> bool {
-        false
+    /// The default implementation uses
+    /// [TransactionManager::is_broken_transaction_manager].
+    fn is_broken(&mut self) -> bool {
+        Self::TransactionManager::is_broken_transaction_manager(self)
     }
 }


weiznich (approved):
:+1: That's exactly what I believed to be already there. 
I will try to cut a 0.3.1 release with that and another recent fix.

PR #100, opened 2023-08-24
Introduce an `AsyncConnectionWrapper` type

### LINKED ISSUE #17: Using the connection for migrations
### Versions

- **Rust:** rustc 1.62.1 (e092d0b6b 2022-07-16)
- **Diesel:**2.0.00-rc.1
- **Diesel_async:** main branch @ [3c9e976](https://github.com/weiznich/diesel_async/tree/3c9e976c1c30d3aa3d0751b89f72b2ce43869c4f)
- **Database:** PostgreSQL
- **Operating System: ** MacOS Monterey

### Feature Flags

- **diesel:** `["postgres_backend", "chrono", "uuid", "numeric", "ipnet-address", "serde_json" ]`, default features inactive
- **diesel_async:** `["postgres"]`
- **diesel_migrations:** `["postgres"]`

## Problem Description

Not really a problem per se, but a missing feature. I would like to not depend on `diesel_cli` to manage migrations, so I was using `diesel_migrations` to handle this, but the `MigrationHarness` needed to retrieve the applied migrations requires the asynchronous connection to implement `MigrationConnection`, which makes sure that the DB has the correct table.

Unfortunately, this is not yet implemented for asynchronous connections.

I'm trying to embed the migrations with my code, and even create the DB if needed, so that on startup, it will check that the DB is at the correct migration.

I would expect to be able to do this with an asynchronous connection, the same way as with a synchronous connection.

```
error[E0277]: the trait bound `AsyncPgConnection: MigrationConnection` is not satisfied
   --> backend/src/lib.rs:152:20
    |
152 |     run_migrations(&mut conn).expect("couldn't run migrations");
    |     -------------- ^^^^^^^^^ the trait `MigrationConnection` is not implemented for `AsyncPgConnection`
    |     |
    |     required by a bound introduced by this call
    |
    = help: the trait `MigrationConnection` is implemented for `PgConnection`
    = note: required because of the requirements on the impl of `MigrationHarness<Pg>` for `AsyncPgConnection`
note: required by a bound in `update_database::{closure#0}::run_migrations`
   --> backend/src/lib.rs:123:31
    |
122 |     fn run_migrations(
    |        -------------- required by a bound in this
123 |         connection: &mut impl MigrationHarness<Pg>,
    |                               ^^^^^^^^^^^^^^^^^^^^ required by this bound in `update_database::{closure#0}::run_migrations`

error[E0277]: the trait bound `AsyncPgConnection: LoadConnection` is not satisfied
   --> backend/src/lib.rs:152:20
    |
152 |     run_migrations(&mut conn).expect("couldn't run migrations");
    |     -------------- ^^^^^^^^^ the trait `LoadConnection` is not implemented for `AsyncPgConnection`
    |     |
    |     required by a bound introduced by this call
    |
    = help: the trait `LoadConnection<B>` is implemented for `PgConnection`
    = note: required because of the requirements on the impl of `diesel::query_dsl::LoadQuery<'_, AsyncPgConnection, MigrationVersion<'static>>` for `SelectStatement<FromClause<diesel_migrations::migration_harness::__diesel_schema_migrations::table>, query_builder::select_clause::SelectClause<diesel_migrations::migration_harness::__diesel_schema_migrations::columns::version>, query_builder::distinct_clause::NoDistinctClause, query_builder::where_clause::NoWhereClause, query_builder::order_clause::OrderClause<diesel::expression::operators::Desc<diesel_migrations::migration_harness::__diesel_schema_migrations::columns::version>>>`
    = note: required because of the requirements on the impl of `MigrationHarness<Pg>` for `AsyncPgConnection`
note: required by a bound in `update_database::{closure#0}::run_migrations`
   --> backend/src/lib.rs:123:31
    |
122 |     fn run_migrations(
    |        -------------- required by a bound in this
123 |         connection: &mut impl MigrationHarness<Pg>,
    |                               ^^^^^^^^^^^^^^^^^^^^ required by this bound in `update_database::{closure#0}::run_migrations`
```

No additional error is shown.

### Steps to reproduce

```rust
    /// Embedded DB migrations.
    const MIGRATIONS: EmbeddedMigrations = embed_migrations!("../migrations");

    /// Runs the migrations in the database.
    #[allow(trivial_casts)]
    fn run_migrations(
        connection: &mut impl MigrationHarness<Pg>,
    ) -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
        let migrations = (&MIGRATIONS as &dyn MigrationSource<Pg>).migrations()?;
        let last = migrations
            .last()
            .expect("could not find the last migration");

        if let Some(last_applied) = connection.applied_migrations()?.first() {
            if last_applied > &last.name().version() {
                // TODO: downgrade, probably add a confirmation parameter in the executable.
                // But how? We would need to get the migrations for downgrade, but we don't have them.
                panic!("Not possible to downgrade the database");
            }
        }

        let _ = connection.run_pending_migrations(MIGRATIONS)?;
        Ok(())
    }

    let mut conn = match db::Connection::establish(db_cfg.conn_str()).await {
        Err(e) => {
            eprintln!("WARNING: error connecting to {}", db_cfg.conn_str());
            db::create(db_cfg)
                .await
                .expect("could not create the database")
        }
        Ok(c) => c,
    };
```

## Checklist

- [ ] This issue can be reproduced without requiring a third party crate

## Suggestion

I can work on this, it should be fairly easy to implement, but, should I add a feature to add this migrations connection implementation, so that it's not active by default for those who don't want to use migrations?

weiznich:
This type turns a `diesel_async::AsyncConnection` into a `diesel::Conenction`. I see the following use-cases for this:

* Having a pure rust sync diesel connection implementation for postgres and mysql can simplify the setup of new diesel projects
* Allowing projects depending on `diesel_async` to use `diesel_migrations` without depending on `libpq`/`libmysqlclient`

This change requires restructuring the implementation of `AsyncPgConnection` a bit so that we make the returned future `Send` independently of whether or not the query parameters are `Send`. This is possible by serialising the bind parameters before actually constructing the future.

It also refactors the `TransactionManager` implementation to share more code with diesel itself.

Fixes #17

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All user visible changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/), as described
 for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/text/1105-api-evolution.md)
 
+## Unreleased
+
+* Add a `AsyncConnectionWrapper` type to turn a `diesel_async::AsyncConnection` into a `diesel::Connection`. This might be used to execute migrations via `diesel_migrations`. 
+
 ## [0.3.2] - 2023-07-24
 
 * Fix `TinyInt` serialization
@@ -52,5 +56,3 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 [0.2.1]: https://github.com/weiznich/diesel_async/compare/v0.2.0...v0.2.1
 [0.2.2]: https://github.com/weiznich/diesel_async/compare/v0.2.1...v0.2.2
 [0.3.0]: https://github.com/weiznich/diesel_async/compare/v0.2.0...v0.3.0
-[0.3.1]: https://github.com/weiznich/diesel_async/compare/v0.3.0...v0.3.1
-[0.3.2]: https://github.com/weiznich/diesel_async/compare/v0.3.1...v0.3.2
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,11 +13,11 @@ description = "An async extension for Diesel the safe, extensible ORM and Query 
 rust-version = "1.65.0"
 
 [dependencies]
-diesel = { version = "~2.1.0", default-features = false, features = ["i-implement-a-third-party-backend-and-opt-into-breaking-changes"]}
+diesel = { version = "~2.1.1", default-features = false, features = ["i-implement-a-third-party-backend-and-opt-into-breaking-changes"]}
 async-trait = "0.1.66"
 futures-channel = { version = "0.3.17", default-features = false, features = ["std", "sink"], optional = true }
 futures-util = { version = "0.3.17", default-features = false, features = ["std", "sink"] }
-tokio-postgres = { version = "0.7.2", optional = true}
+tokio-postgres = { version = "0.7.10", optional = true}
 tokio = { version = "1.26", optional = true}
 mysql_async = { version = ">=0.30.0,<0.33", optional = true}
 mysql_common = {version = ">=0.29.0,<0.31.0", optional = true}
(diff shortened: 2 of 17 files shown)


iMobs:
Is there going to be an example or docs on how to use this with `diesel_migrations`?

weiznich:
@iMobs I'm happy to accept your contribution for such examples or documentation. Otherwise this is a currently unreleased feature, therefore its expected that the documentation is incomplete.

Issue #104, opened 2023-09-04
Compiler requires `'static` when using transactions on generic AsyncConnection


kmdreko:
The following reproducible sample triggers a compiler error:

```
[dependencies]
diesel = { version = "2.1.1", features = ["postgres"] }
diesel-async = { version = "0.4.1", features = ["postgres", "bb8"] }
```

```
use diesel::pg::Pg;
use diesel::result::Error;
use diesel_async::scoped_futures::ScopedFutureExt;
use diesel_async::{AsyncConnection, RunQueryDsl};

pub async fn demo<T: AsyncConnection<Backend = Pg>>(conn: &mut T) -> Result<(), Error> {
    conn.transaction::<(), Error, _>(|conn| async {
            diesel::sql_query("insert into posts (title, body) values ('title', 'body')").execute(conn).await?;
            Ok(())
        }.scope_boxed()).await?;
    Ok(())
}
```

```
error[E0310]: the parameter type `T` may not live long enough
  --> src/main.rs:7:45
   |
7  |       conn.transaction::<(), Error, _>(|conn| async {
   |  _____________________________________________^
8  | |             diesel::sql_query("insert into posts (title, body) values ('title', 'body')").execute(conn).await?;
9  | |             Ok(())
10 | |         }.scope_boxed()).await?;
   | |_______________________^ ...so that the type `T` will meet its required lifetime bounds
   |
help: consider adding an explicit lifetime bound...
   |
6  | pub async fn demo<T: AsyncConnection<Backend = Pg> + 'static>(conn: &mut T) -> Result<(), Error> {
   |                                                    +++++++++
```

- [x] I have already looked over the [issue tracker](https://github.com/weiznich/diesel_async/issues) for similar possible closed issues.
- [x] This issue can be reproduced on Rust's stable channel. (Your issue will be closed if this is not the case)
- [x] This issue can be reproduced without requiring a third party crate

---

As far as I know, using `.transaction()` and `.execute()` in this way should not require a `'static` bound.

I ended up investigating myself, created [a StackOverflow post](https://stackoverflow.com/questions/77034609/why-does-this-combination-of-gats-lifetimes-and-async-require-t-static) about it when I thought I was stuck, and ended up self-answering it as an odd interaction with GATs and `Self`-bounds.

[...]

PR #106, opened 2023-09-05
fix: avoid where Self: 'a clause on GATs

kmdreko:
This removes the `where Self: 'a` clauses on `AsyncConnection`'s associated types. The compiler over-eagerly asserts the trait should have the constraint but it is flawed when handling transactions generically. Dummy `_silence_lint_on_*` functions were added to convince the compiler the constraints are not required.

Fixes: #104

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -127,21 +127,13 @@ pub trait SimpleAsyncConnection {
 #[async_trait::async_trait]
 pub trait AsyncConnection: SimpleAsyncConnection + Sized + Send {
     /// The future returned by `AsyncConnection::execute`
-    type ExecuteFuture<'conn, 'query>: Future<Output = QueryResult<usize>> + Send
-    where
-        Self: 'conn;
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -341,4 +333,10 @@ pub trait AsyncConnection: SimpleAsyncConnection + Sized + Send {
     fn transaction_state(
         &mut self,
     ) -> &mut <Self::TransactionManager as TransactionManager<Self>>::TransactionStateData;
+
+    #[doc(hidden)]
+    fn _silence_lint_on_execute_future(_: Self::ExecuteFuture<'_, '_>) {}
```
weiznich, line 338: Can you add a comment to these function explaining why they exist + linking to the relevant rust issues?

weiznich (commented):
I'm in favor of this change, but I would like to add some more documentation to the workaround so that we don't wonder why this was done.

weiznich (approved):
Thanks :+1:

Issue #115, opened 2023-09-30
AsyncConnectionWrapper does not implement `build_transaction`


banool:
### Versions

- **Rust:** 1.71.1
- **Diesel:** 2.1.1
- **Diesel_async:** 0.4.1
- **Database:** Postgres
- **Operating System** MacOS 14

### Feature Flags

- **diesel:** postgres_backend, r2d2, serde_json, numeric, chrono
- **diesel_async:** postgres, async-connection-wrapper, r2d2, tokio

## Problem Description
I have the following types defined:
```
use diesel::{
    pg::Pg,
    query_builder::{AstPass, Query, QueryFragment},
    r2d2::{ConnectionManager, PoolError, PooledConnection},
    QueryResult, RunQueryDsl,
};
use diesel_async::{
    AsyncPgConnection, async_connection_wrapper::AsyncConnectionWrapper,
};
use std::{cmp::min, sync::Arc};

pub type MyDbConnection = AsyncConnectionWrapper<AsyncPgConnection>;
pub type PgPool = diesel::r2d2::Pool<ConnectionManager<MyDbConnection>>;
pub type PgDbPool = Arc<PgPool>;
pub type PgPoolConnection = PooledConnection<ConnectionManager<MyDbConnection>>;
```

and this function to get a pool:

```
pub fn new_db_pool(database_url: &str) -> Result<PgDbPool, PoolError> {
    let manager = ConnectionManager::<MyDbConnection>::new(database_url);
    PgPool::builder().build(manager).map(Arc::new)
}
```

This all compiles fine.

When I try to use my `PgPoolConnection` connection however, like this:
```
fn insert_to_db(
    conn: &mut PgPoolConnection,
) {
    let result = conn.build_transaction()
         .read_write()
        .run::<_, Error, _>(|pg_conn| {
            insert_to_db_impl(
                pg_conn,
                &current_ans_lookups,
                &ans_lookups,
                &current_ans_primary_names,
                &ans_primary_names,
                &current_ans_lookups_v2,
                &ans_lookups_v2,
                &current_ans_primary_names_v2,
                &ans_primary_names_v2,
            )
...
```

I get this error:
```
no method named `build_transaction` found for mutable reference `&mut PooledConnection<ConnectionManager<AsyncConnectionWrapper<AsyncPgConnection, Tokio>>>` in the current scope
```

[...]

PR #116, opened 2023-09-30
Expose build_transaction through AsyncConnectionWrapper

banool:
This addresses https://github.com/weiznich/diesel_async/issues/115.

I feel I've made the section in async_connection_wrapper more complex than necessary, I can try tidy it up.

The tests don't pass right now but I know the core section compiles.

Let me know what you think about this solution!

### DIFF
--- a/src/async_connection_wrapper.rs
+++ b/src/async_connection_wrapper.rs
@@ -200,6 +200,17 @@ mod implementation {
         }
     }
 
+    #[cfg(feature = "postgres")]
+    impl<C, B> crate::pg::BuildTransaction<C> for AsyncConnectionWrapper<C, B>
+    where
+        C: crate::AsyncConnection + crate::pg::BuildTransaction<C>,
+        B: BlockOn + Send,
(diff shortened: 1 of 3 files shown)


banool:
Another approach would be to add a function that lets the user call functions inner.

weiznich:
Thanks for opening this PR. It addresses the problem that there is no `build_transaction` function for the async connection wrapper type, but I don't feel that the implemented solution is particular help full for down stream crates using that type because the indirectly exposed `run` function is still async.  I would prefer a solution that exposes a sync run method as well. That migth require duplicating or at least modifying parts of `TransactionBuilder` type, by having a separate `run` function depending on whether the inner connection type implements `diesel::Connection` or `diesel_async::AsyncConnection`.

banool:
Hmmmm yeah good point. In my case I'm actually already in an async environment so I think I'll just try switch to using diesel-async proper, so I'm going to just close this PR for now. I'll leave the issue open though.

PR #121, opened 2023-10-05
Expose underlying connection errors to user of AsyncPgConnection

banool:
## Summary
This PR addresses https://github.com/weiznich/diesel_async/pull/118. In short, rather than just printing when there is an error with the tokio postgres connection, we expose it to the user in query path. If there is an error, they will get that rather than the inner result.

## Test Plan
Set up DB:
```
psql postgres://dport@127.0.0.1:5432/postgres -c 'DROP DATABASE IF EXISTS diesel_test'
psql postgres://dport@127.0.0.1:5432/postgres -c 'CREATE DATABASE diesel_test'
export DATABASE_URL='postgres://dport:@127.0.0.1:5432/diesel_test'
```

Run tests:
```
cargo test --features postgres
```

I don't have any new tests yet. Adding a test for this behavior would be quite hard, we'd have to introduce some kind of induced failure + a way to have a deterministic order of steps. We can discuss here after we decide if the core change is good.

### DIFF
--- a/src/pg/mod.rs
+++ b/src/pg/mod.rs
@@ -21,6 +21,7 @@ use futures_util::future::BoxFuture;
 use futures_util::stream::{BoxStream, TryStreamExt};
 use futures_util::{Future, FutureExt, StreamExt};
 use std::borrow::Cow;
+use std::ops::DerefMut;
 use std::sync::Arc;
 use tokio::sync::Mutex;
 use tokio_postgres::types::ToSql;
@@ -102,6 +103,7 @@ pub struct AsyncPgConnection {
     stmt_cache: Arc<Mutex<StmtCache<diesel::pg::Pg, Statement>>>,
     transaction_state: Arc<Mutex<AnsiTransactionManager>>,
     metadata_cache: Arc<Mutex<PgMetadataCache>>,
+    error_receiver: Arc<Mutex<tokio::sync::oneshot::Receiver<diesel::result::Error>>>,
 }
 
 #[async_trait::async_trait]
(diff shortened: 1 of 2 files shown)


weiznich (changes requested):
Thanks for opening this PR :+1: It's more or less what I had in mind.

I've left a few comments about things that I would like to change. In addition we might need to make sure that we also perform that selection in `SimpleConnection::batch_execute`.

src/pg/mod.rs
```diff
@@ -411,8 +440,12 @@ impl AsyncPgConnection {
             let res = callback(raw_connection, stmt.clone(), binds).await;
             let mut tm = tm.lock().await;
             update_transaction_manager_status(res, &mut tm)
+        };
+
+        match self.error_receiver.try_recv() {
```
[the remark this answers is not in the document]
banool: I tried that but `tokio::select` only works in an async context. To deal with that I tried to move it into the above async closure but I had difficulty with that too. Let me try the latter again.

banool:
Hey hey, just got this working!

At first I tried this: https://gist.github.com/banool/b62cda55a4cac09c0d57e6b283b40778. This doesn't work because `self` is captured by the returned future. I can't clone the error receiver like we do with everything else that we feed into the future because it is not Clone.

At first I was relucant to use Arc + Mutex to solve this but I see you're doing this already in other cases. So that's what I did.

Check out the new code!

banool:
If this looks good I can make that `batch_execute` change you suggested too.

src/pg/mod.rs
```diff
@@ -102,6 +103,7 @@ pub struct AsyncPgConnection {
     stmt_cache: Arc<Mutex<StmtCache<diesel::pg::Pg, Statement>>>,
     transaction_state: Arc<Mutex<AnsiTransactionManager>>,
     metadata_cache: Arc<Mutex<PgMetadataCache>>,
+    error_receiver: Arc<Mutex<tokio::sync::oneshot::Receiver<diesel::result::Error>>>,
```
weiznich, line 106: I'm not sure if using a mutex here is a good idea, because that one would be locked while executing the query (compared to while starting to prepare as for the other mutexes). That would basically restrict the ability to do pipe lining at all. I need to think about a better solution, probably using a `mpmc` channel or something like that.

banool: Perhaps we could make it only take the mutex as part of the select arm so it doesn't lock the whole time. Not sure if that works though.

banool: Hmm actually that doesn't really work yeah, you need to take the lock to await the channel as a future. Perhaps some approach where we use try_recv and if it returns Ok, we return a completed future. I tried to code this up but it didn't really work.

I notice that some other tokio channels have maybe what we need, this poll_recv function: https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Receiver.html#method.poll_recv.

weiznich: I've finally found the time to have another look at this. #132 is what I've settled on.

banool: Neato, looks good to me! I'll close this one out.

PR #134, opened 2023-11-29
enable rustls with mysql

Ludea:
Creating a new feature to enable rustls support with mysql


Ludea:
adding crate directly and enable this feature does the jobs

PR #140, opened 2024-03-01
Fix #138 (Incoherent UPDATE return for MySQL/MariaDB)

### LINKED ISSUE #138: Incoherence between Diesel and Diesel_async on MySQL/MariaDB connection (UPDATE return value)
### Versions

- **Rust:** 1.76
- **Diesel:** 2.1.4
- **Diesel_async:** master (0.4.1)
- **Database:** MariaDB 10.16
- **Operating System:** Debian testing

### Feature Flags

- **diesel:** mysql, chrono
- **diesel_async:** mysql, bb8

## Problem Description

I'm sorry in advance if this issue is badly reported, I'm a bit new to this...

Note: In this issue, MariaDB is used to describe both MariaDB & MySQL, as I think both are affected, but I only tested MariaDB.

In current Diesel version, a MySQL/MariaDB connection will see a flag activated to change the way UPDATE operations work:

https://github.com/diesel-rs/diesel/blob/ad8ef479e1c6f926b940ad9a43877fc5a993ec5f/diesel/src/mysql/connection/url.rs#L134

```rs
        // this is not present in the database_url, using a default value
        let client_flags = CapabilityFlags::CLIENT_FOUND_ROWS;
```

On PostgreSQL, UPDATE operations will return the number of lines matched by the operation, even if the values set are identical to the ones already present in a row.
On MariaDB/MySQL, only the number of **changed** rows is reported by default. (Relevant StackOverflow answer: https://stackoverflow.com/a/2186952; Extract from MariaDB manual: https://dev.mysql.com/doc/c-api/8.3/en/mysql-affected-rows.html )

On Diesel, to get a consistent behavior between both engines, it seems like the flag to enable reporting of *matching* rows rather than *modified* rows is set on by default.

As Diesel-async seems to use an other way to parse the database URL, the same option isn't set by default and must be specified in the URL.
If we (as I got helped by someone to read diesel-async codebase) got the code right, diesel-async uses mysql-async, which supports setting this option:
https://docs.rs/mysql_async/0.32.2/mysql_async/struct.Opts.html#method.client_found_rows

I expected that the option would also be set in diesel-async to get a coherent and consistent behavior when switching between Diesel and Diesel-async.
In my case, I had tests which relied on reporting affected rows, which had to be worked-around when implementing support for MariaDB.

It would be nice for either diesel-async to also enable that setting by default, or for diesel to actually implement similar parsing of URLs to allow enabling/disabling the option using the URL.

UPDATE commands should either return changed rows count, or matching rows count, similarly regardless of if you use diesel or diesel-async

UPDATE commands only report changed rows in diesel-async, while it reports matching rows in diesel

No

### Steps to reproduce

I don't have a MWE for now, but it shouldn't be too hard to produce one if needed

1. Create a MariaDB database, with matching DB & user
2. Create a basic table with a key and a value (like an integer) (for example, id: u32 and val: u64)
3. Insert a value to this database ( for example, id=1 and val=42 )
4. Update the values to the same value ( so here, update for id=1 values val=42 )
5. The update command will return "1 row affected" on Diesel, and "0 row affected" on Diesel-Async

## Checklist

Soblow:
This MR fixes #138 , and also adds a few missing links in CHANGELOG.md .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 ## [Unreleased]
 
 * Added type `diesel_async::pooled_connection::mobc::PooledConnection`
+* MySQL/MariaDB now use `CLIENT_FOUND_ROWS` capability to allow consistent behavior with PostgreSQL regarding return value of UPDATe commands.
 
 ## [0.4.1] - 2023-09-01
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,3 +69,6 @@ in the pool should be checked if they are still valid
 [0.3.0]: https://github.com/weiznich/diesel_async/compare/v0.2.0...v0.3.0
 [0.3.1]: https://github.com/weiznich/diesel_async/compare/v0.3.0...v0.3.1
 [0.3.2]: https://github.com/weiznich/diesel_async/compare/v0.3.1...v0.3.2
+[0.4.0]: https://github.com/weiznich/diesel_async/compare/v0.3.2...v0.4.0
+[0.4.1]: https://github.com/weiznich/diesel_async/compare/v0.4.0...v0.4.1
+[Unreleased]: https://github.com/weiznich/diesel_async/compare/v0.4.1...main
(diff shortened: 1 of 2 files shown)


Soblow:
I can add some tests if you want though.

weiznich (approved):
Thanks :+1: 

Having a test for this would be great (but it's not strictly required for merging)

PR #146, opened 2024-04-08
Add SyncConnectionWrapper type

momobel:
Handles https://github.com/weiznich/diesel_async/discussions/95 

This type wraps a `diesel::connection::Connection` fulfilling needed
requirement with a `diesel_async::AsyncConnection` trait.

It can be useful  when one desires
* using a sync `Connection` implementation (sqlite) in async context
* using the same code base within async crates needing multiple backends
  (sqlite + postgres)

It additionally fixes 2 small issues (first 2 commits):
- missing instrumentation (needed for latest diesel version)
- add missing tokio feature dep

Checklist before merging
- [x] Update Cargo.toml to depend on version of diesel with required changes (either git master or tagged version)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,8 @@ diesel_migrations = "2.1.0"
 default = []
 mysql = ["diesel/mysql_backend", "mysql_async", "mysql_common", "futures-channel", "tokio"]
 postgres = ["diesel/postgres_backend", "tokio-postgres", "tokio", "tokio/rt"]
+sqlite = ["diesel/sqlite", "sync-connection-wrapper"]
+sync-connection-wrapper = ["tokio/rt"]
 async-connection-wrapper = ["tokio/net"]
 r2d2 = ["diesel/r2d2"]
 
@@ -57,4 +59,5 @@ members = [
         ".",
         "examples/postgres/pooled-with-rustls",
         "examples/postgres/run-pending-migrations-with-rustls",
+        "examples/sync-wrapper",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,7 @@ diesel_migrations = "2.1.0"
 default = []
 mysql = ["diesel/mysql_backend", "mysql_async", "mysql_common", "futures-channel", "tokio"]
 postgres = ["diesel/postgres_backend", "tokio-postgres", "tokio", "tokio/rt"]
-async-connection-wrapper = []
+async-connection-wrapper = ["tokio/net"]
 r2d2 = ["diesel/r2d2"]
 
 [[test]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,3 +61,6 @@ members = [
         "examples/postgres/run-pending-migrations-with-rustls",
         "examples/sync-wrapper",
 ]
+
+[patch.crates-io]
+diesel = { git = "http://github.com/diesel-rs/diesel", rev = "793de72" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,7 +49,7 @@ path = "tests/lib.rs"
 harness = true
 
 [package.metadata.docs.rs]
-features = ["postgres", "mysql", "deadpool", "bb8", "mobc", "async-connection-wrapper", "r2d2"]
+features = ["postgres", "mysql", "sqlite", "deadpool", "bb8", "mobc", "async-connection-wrapper", "sync-connection-wrapper", "r2d2"]
 no-default-features = true
 rustc-args = ["--cfg", "doc_cfg"]
 rustdoc-args = ["--cfg", "doc_cfg"]
--- a/examples/sync-wrapper/Cargo.toml
+++ b/examples/sync-wrapper/Cargo.toml
@@ -1,0 +1,17 @@
+[package]
+name = "sync-wrapper"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+diesel = { version = "2.1.0", default-features = false }
+diesel-async = { version = "0.4.0", path = "../../", features = ["sync-connection-wrapper", "async-connection-wrapper"] }
+diesel_migrations = "2.1.0"
+futures-util = "0.3.21"
+tokio = { version = "1.2.0", default-features = false, features = ["macros", "rt-multi-thread"] }
+
+[features]
+default = ["sqlite"]
+sqlite = ["diesel-async/sqlite"]
--- a/examples/sync-wrapper/src/main.rs
+++ b/examples/sync-wrapper/src/main.rs
@@ -1,0 +1,88 @@
+use diesel::prelude::*;
+use diesel::sqlite::{Sqlite, SqliteConnection};
+use diesel_async::async_connection_wrapper::AsyncConnectionWrapper;
+use diesel_async::sync_connection_wrapper::SyncConnectionWrapper;
+use diesel_async::{AsyncConnection, RunQueryDsl, SimpleAsyncConnection};
+use diesel_migrations::{embed_migrations, EmbeddedMigrations, MigrationHarness};
+
+// ordinary diesel model setup
+
+table! {
+    users {
+        id -> Integer,
+        name -> Text,
+    }
+}
+
+#[derive(Debug, Queryable, Selectable)]
+#[diesel(table_name = users)]
+struct User {
+    id: i32,
+    name: String,
+}
+
+const MIGRATIONS: EmbeddedMigrations = embed_migrations!();
+
+type InnerConnection = SqliteConnection;
+
+type InnerDB = Sqlite;
+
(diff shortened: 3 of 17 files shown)


Cargo.toml
```diff
@@ -57,4 +59,8 @@ members = [
         ".",
         "examples/postgres/pooled-with-rustls",
         "examples/postgres/run-pending-migrations-with-rustls",
+        "examples/generic-async",
 ]
+
+[patch.crates-io]
+diesel = { path = "../diesel/diesel" }
```
momobel, line 66: Note to self: change this once diesel is updated with needed changes or released with them

weiznich: For now we can also use a git dependency to your branch here. That we at least the CI can run and test something.

src/sync_connection_wrapper.rs
```diff
@@ -0,0 +1,254 @@
...
+    )
+}
+
+/// A wrapper of a [`diesel::connection::Connection`] usable in async context.
+///
+/// It implements AsyncConnection if [`diesel::connection::Connection`] fullfils requirements:
+/// * it's a [`diesel::connection::LoadConnection`]
+/// * its [`diesel::connection::Connection::Backend`] has a [`diesel::query_builder::BindCollector`] implementing [`diesel::query_builder::MoveableBindCollector`]
+/// * its [`diesel::connection::LoadConnection::Row`] implements [`diesel::row::IntoOwnedRow`]
+pub struct SyncConnectionWrapper<C> {
```
weiznich, line 40: It would be great to have an example here similar to what we have for the `AsyncConnectionWrapper`.

We also might want to mention that for now only sqlite is supported.

src/sync_connection_wrapper.rs
```diff
@@ -0,0 +1,254 @@
...
+    pub fn new(connection: C) -> Self
+    where
+        C: Connection,
+    {
+        SyncConnectionWrapper {
+            inner: Arc::new(Mutex::new(connection)),
+        }
+    }
+
+    pub(self) fn spawn_blocking<'a, R>(
```
weiznich, line 167: Mostly a note for myself: I want to abstract that later over different `spawn_blocking` functions, so that you could use different runtimes here.

(That's nothing that needs to be addressed in this PR)

weiznich (commented):
Thanks for opening this PR. This looks almost good. I've only spotted a few minor suggestions related to documentation, otherwise it seems to be ready to be merged.

I've pushed a commit here (https://github.com/weiznich/diesel/tree/sync_connection_wrapper_benches) that hooks up everything in `diesel_benches`.  It seems like this implementation is roughly 5-6 times slower than the optimized sync diesel variant. On the other hand, for queries it's already faster than sea-orm and at least as fast as sqlx (for larger response type sizes it's drastically faster). So good work on that. (Inserts are slower due to the missing support for batch inserts)

Cargo.toml
momobel: Added in a specific commit so it's easier to remove

examples/sync-wrapper/src/main.rs
```diff
@@ -0,0 +1,88 @@
...
+}
+
+const MIGRATIONS: EmbeddedMigrations = embed_migrations!();
+
+type InnerConnection = SqliteConnection;
+
+type InnerDB = Sqlite;
+
+async fn establish(db_url: &str) -> ConnectionResult<SyncConnectionWrapper<InnerConnection>> {
+    SyncConnectionWrapper::<SqliteConnection>::establish(db_url).await
```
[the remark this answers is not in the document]
momobel: Handled

src/sync_connection_wrapper.rs
momobel: Implemented and handled it in DSL doc examples as well.

src/sync_connection_wrapper.rs
momobel: That would be a task I am interested in after this PR!

momobel:
> I've pushed a commit here (https://github.com/weiznich/diesel/tree/sync_connection_wrapper_benches) that hooks up everything in `diesel_benches`. It seems like this implementation is roughly 5-6 times slower than the optimized sync diesel variant. On the other hand, for queries it's already faster than sea-orm and at least as fast as sqlx (for larger response type sizes it's drastically faster). So good work on that. (Inserts are slower due to the missing support for batch inserts)

Great, thanks for that.
That's a lot slower but it works at least. For sure this could need optimizations afterwards.

weiznich:
> That's a lot slower but it works at least. For sure this could need optimizations afterwards.

It's not that unexpected that it's slower than the sync diesel version. I've spend quite a bit on time optimizing that one over the years. 
I would word it the order way around: It's good that this is already, without any optimizations, as fast as the other async implementations in the ecosystem.

That written: I had a quick go at the code and implemented a simple optimization that just caches the column names for rows of the same query, so that we don't allocate them over and over again. This improves the performance of the benchmarks that loads 10.000 rows in one query by ~20-30%, which then makes it only 2-3 times slower than sync diesel. The cost is that it's a tiny bit slower for the case where we only load one row, due to the caching overhead. You can find this changes here (https://github.com/diesel-rs/diesel/commit/97e34d4dbccc5e0856894c42b6eb9c62cccb0020) and here (https://github.com/weiznich/diesel_async/commit/d415a2d2455c3a6829227c26961204e52821d02b). Might be worth to just pull them in as well, although we likely want to make that `column_name_cache` type an associated type on the `OwnedRow` trait so that other backends could chose different cache types there.

Cargo.toml
```diff
@@ -38,7 +38,9 @@ diesel_migrations = "2.1.0"
 default = []
 mysql = ["diesel/mysql_backend", "mysql_async", "mysql_common", "futures-channel", "tokio"]
 postgres = ["diesel/postgres_backend", "tokio-postgres", "tokio", "tokio/rt"]
-async-connection-wrapper = []
+sqlite = ["diesel/sqlite", "sync-connection-wrapper"]
+sync-connection-wrapper = ["tokio/rt"]
+async-connection-wrapper = ["tokio/net"]
```
weiznich, line 43:
```suggestion
async-connection-wrapper = ["tokio/rt"]
```

should be enough for that feature, right?

Cargo.toml
```diff
@@ -47,7 +49,7 @@ path = "tests/lib.rs"
 harness = true
 
 [package.metadata.docs.rs]
-features = ["postgres", "mysql", "deadpool", "bb8", "mobc", "async-connection-wrapper", "r2d2"]
+features = ["postgres", "mysql", "sqlite", "deadpool", "bb8", "mobc", "async-connection-wrapper", "r2d2"]
```
weiznich, line 52:
```suggestion
features = ["postgres", "mysql", "sqlite", "deadpool", "bb8", "mobc", "async-connection-wrapper", "sync-connection-wrapper", "r2d2"]
```

examples/sync-wrapper/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+[dependencies]
+diesel = { version = "2.1.0", default-features = false }
+diesel-async = { version = "0.4.0", path = "../../", features = ["sync-connection-wrapper", "async-connection-wrapper"] }
+diesel_migrations = "2.1.0"
+futures-util = "0.3.21"
+tokio = { version = "1.2.0", default-features = false, features = ["macros", "rt-multi-thread"] }
+
+[features]
+default = ["sqlite"]
+sqlite = ["diesel/sqlite"]
```
weiznich, line 17:
```suggestion
sqlite = ["diesel-async/sqlite"]
```

src/doctest_setup.rs
```diff
@@ -160,6 +160,81 @@ cfg_if::cfg_if! {
...
+    }  else if #[cfg(feature = "sqlite")] {
+        use diesel_async::sync_connection_wrapper::SyncConnectionWrapper;
+        use diesel::sqlite::SqliteConnection;
+        #[allow(dead_code)]
+        type DB = diesel::sqlite::Sqlite;
+        #[allow(dead_code)]
+        type DbConnection = SyncConnectionWrapper<SqliteConnection>;
+
+        fn database_url() -> String {
+            database_url_from_env("SQLITE_UNIT_TEST_DATABASE_URL")
```
weiznich, line 174:
```suggestion
            database_url_from_env("SQLITE_DATABASE_URL")
```

(As we don't have the same concurrent schema modification issues as with mysql here)

src/sync_connection_wrapper.rs
```diff
@@ -0,0 +1,286 @@
+//! This module contains a wrapper type
+//! that provides a [`crate::AsyncConnection`]
+//! implementation for types that implement
+//! [`diesel::connection`]. Using this type
```
weiznich, line 4:
```suggestion
//! [`diesel::Connection`]. Using this type
```

src/sync_connection_wrapper.rs
weiznich: That would be great :pray:

weiznich (commented):
This looks good as well beside some minor notes about documentation and feature flags. 

As for the CI failures: We need to replace any usage of `diesel::sql_function!` with `diesel::define_sql_function!`, after this it should pass.

weiznich (approved):

Issue #103, opened 2023-09-01
Failed to find a type oid for ENUM_TYPE


omid:
### Versions

- **Rust:** 1.72
- **Diesel:** 2.1
- **Diesel_async:** 0.4.1
- **Database:** postgresql
- **Operating System** linux

## Problem Description
After I upgrade the code from diesel-async 0.3.x to 0.4, I faced this error: `Failed to find a type oid for ENUM_TYPE`.
After I downgrade, the error disappeared. The only change in the code base was diesel-async version.

weiznich:
Can you please provide an reproducible example, otherwise this report is not actionable.

omid:
I tested it on a big project. 
I'll try to make a minimum code base next week 🙏

omid:
@weiznich, thanks. The problem is in `eq_any` (`eq` is fine). 
Within `custom_types_round_trip` test, this following code fails:
```
custom_types::table
        .filter(custom_types::custom_enum.eq_any([MyEnum::Foo]))
        .load(connection)
        .await
        .unwrap();
```

with this error:
```
thread 'custom_types::custom_types_round_trip' panicked at 'called `Result::unwrap()` on an `Err` value: SerializationError(FailedToLookupTypeError(PgMetadataCacheKey { schema: None, type_name: "my_type" }))'
```

weiznich:
Thanks for providing this example. I can reproduce the issue with that.

neogenie:
I confirm the bug is reproducible

weiznich:
The problematic code here is that one:

https://github.com/weiznich/diesel_async/blob/ec38eca3d7fccf4097fae20d655a26f11df015ee/src/pg/mod.rs#L319-L382

For custom types we need to resolve the type name to a oid, which might require executing a database query. I've choosen to delay resolving these oid's from the types names until we collected all bind parameters, but this has an important problem: Sometimes this oid is used during bind value serialization (for example for arrays: https://github.com/diesel-rs/diesel/blob/ac0b075f9ce297e06e8e728fce1573e7e1d8b797/diesel/src/pg/types/array.rs#L102) and then we don't get the serialized bind value at all, but an error message.

Now it's somewhat problematic to fix that because on the one hand we want to collect the bind parameters before entering the `async` below, to prefend a `Send` bound on the query generic type. On the other hand, we cannot execute queries outside an async block as executing a query requires resolving a future. I'm happy for suggestions on how to resolve that situation.

[...]

PR #152, opened 2024-05-31
Bump diesel to 2.2

weiznich:


McDic:
Just curious, is there any plan to officially maintain this crate from `diesel.rs` organization?

Swaagie:
For what it is worth was able to verify this branch works with 2.2 and limited use of geometry features.

weiznich:
@McDic I still consider this not on par with diesel itself due to missing language features. See [this post](https://blog.weiznich.de/blog/async-transaction-problem/) for a long version of the underlying problem. I would like to keep that separate as long as the language level problem is not resolved to avoid people thinking that this is a perfect solution. (To be clear hear that problem also appears for all the other async interfaces.)

dullbananas:
I want this change to be released soon

weiznich:
@dullbananas We all want to do that, but there are still blocking issues (like #103 ) that need to be resolved first. Just requesting a release won't make that happen faster, but instead it will take time away that could be spend on working on these issues. Your contributions are welcome there.

PR #153, opened 2024-06-08
Reduce amount of code in AsyncPgConnection functions that have query type as generic parameter

dullbananas:
https://nnethercote.github.io/perf-book/compile-times.html#llvm-ir

### DIFF
--- a/src/pg/mod.rs
+++ b/src/pg/mod.rs
@@ -403,7 +403,7 @@ impl AsyncPgConnection {
         to_sql_result: QueryResult<()>,
         collect_bind_result: QueryResult<()>,
         query_builder: PgQueryBuilder,
-        bind_collector: RawBytesBindCollector<Pg>,
+        mut bind_collector: RawBytesBindCollector<Pg>,
         metadata_lookup: PgAsyncMetadataLookup,
     ) -> BoxFuture<'a, QueryResult<R>>
     where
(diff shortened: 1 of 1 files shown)


weiznich:
Thanks for opening this PR. Do you have any hard numbers that this change actually improves compile times is any measurable way? Or is it just a general optimization that you would like to apply?

dullbananas:
The total amount of LLVM lines in the lemmy_db_schema crate reduced by 34%

src/pg/mod.rs
```diff
@@ -368,16 +381,42 @@ impl AsyncPgConnection {
...
+            query.collect_binds(&mut bind_collector, &mut metadata_lookup, &Pg),
+            query_builder,
+            bind_collector,
+            metadata_lookup,
+        )
+    }
+
+    fn with_prepared_statement_after_sql_built<'a, F, R>(
+        &mut self,
+        callback: impl FnOnce(Arc<tokio_postgres::Client>, Statement, Vec<ToSqlHelper>) -> F + Send + 'a,
```
weiznich, line 400: It might be possible to reduce the number of generic function a bit further by using the following code here:

```suggestion
        callback: fn(Arc<tokio_postgres::Client>, Statement, Vec<ToSqlHelper>) -> F,
```

The same applies to the generic `with_prepared_statement` function above.

weiznich (approved):
Thanks for clarifying. That's helpful.

I'm conceptually fine with this change. I've noted an location where an additional improvement might be possible.

weiznich (approved):

PR #154, opened 2024-06-08
Allow custom OID serialization

### LINKED ISSUE #103: Failed to find a type oid for ENUM_TYPE

dullbananas:
Fixes #103

### DIFF
--- a/src/pg/mod.rs
+++ b/src/pg/mod.rs
@@ -385,7 +385,7 @@ impl AsyncPgConnection {
             .flat_map(|(bind_index, (bytes_0, bytes_1))| {
                 std::iter::zip(bytes_0.unwrap_or_default(), bytes_1.unwrap_or_default())
                     .enumerate()
-                    .filter_map(|(byte_index, bytes)| (bytes == (0, 1)).then_some((bind_index, byte_index)))
+                    .filter_map(move |(byte_index, bytes)| (bytes == (0, 1)).then_some((bind_index, byte_index)))
             })
             // Avoid storing the bind collectors in the returned Future
             .collect::<Vec<_>>();
@@ -407,9 +407,9 @@ impl AsyncPgConnection {
             // to borther with that at all
             if !metadata_lookup.unresolved_types.is_empty() {
                 let metadata_cache = &mut *metadata_cache.lock().await;
-                let real_oids = HashMap::<u32, u32>::new();
+                let mut real_oids = HashMap::<u32, u32>::new();
 
-                for (index, (ref schema, ref lookup_type_name)) in metadata_lookup.unresolved_types.into_iter().enumerate() {
+                for (index, (schema, lookup_type_name)) in metadata_lookup.unresolved_types.iter().enumerate() {
                     // for each unresolved item
                     // we check whether it's arleady in the cache
                     // or perform a lookup and insert it into the cache
--- a/src/pg/mod.rs
+++ b/src/pg/mod.rs
@@ -24,6 +24,7 @@ use futures_util::stream::{BoxStream, TryStreamExt};
 use futures_util::TryFutureExt;
 use futures_util::{Future, FutureExt, StreamExt};
 use std::borrow::Cow;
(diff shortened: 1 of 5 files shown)


src/pg/mod.rs
```diff
@@ -452,27 +503,53 @@ impl AsyncPgConnection {
...
             .push((schema.map(ToOwned::to_owned), type_name.to_owned()));
-        PgTypeMetadata::from_result(Err(FailedToLookupTypeError::new(cache_key)))
+        PgTypeMetadata::from_result(Ok(self.fake_oids(index)))
+    }
+}
+
+/// Allows unambiguously determining:
+/// * where OIDs are written in `bind_collector.binds` after being returned by `lookup_type`
+/// * determining the maximum hardcoded OID in `bind_collector.metadata`
+struct SameOidEveryTime {
```
weiznich, line 545: I wonder whether we can add a flag to this type that tracks whether or not we hit a custom type (called the lookup function). 
This would allow us to use this knowledge above to skip the second run of the bind collector and the comparison for the common case that we don't have any custom type.

src/pg/mod.rs
```diff
@@ -452,27 +503,53 @@ impl AsyncPgConnection {
...
+    first_byte: u8,
```
weiznich, line 546: I likely would just use a 'u32' directly here instead of constructing it everytime in the lookup function.

src/pg/mod.rs
```diff
@@ -367,10 +366,35 @@ impl AsyncPgConnection {
...
+            &mut bind_collector_1,
+            &mut SameOidEveryTime { first_byte: 1 },
+            &Pg,
+        );
+
+        let mut metadata_lookup = PgAsyncMetadataLookup::new(&bind_collector_0);
         let collect_bind_result =
             query.collect_binds(&mut bind_collector, &mut metadata_lookup, &Pg);
 
+        let fake_oid_locations = std::iter::zip(bind_collector_0.binds, bind_collector_1.binds)
```
weiznich, line 387: I'm not sure if it might be worth to do that in the future anyway just do allow us to perform the comparison and the replacement in one pass.

src/pg/mod.rs
weiznich: Another alternative would be to just collect the relevant type names here so that we already have them for later lookups. And we can check if that list is empty before doing the expensive stuff. 

The other quick idea that I have is to use something uniquely identifiable per type instead a constant OID. Maybe a truncated hash value of the type name? (For the second variant we can just add a constant, because only the difference in the output is relevant, right?)

src/pg/mod.rs
```diff
@@ -367,10 +366,35 @@ impl AsyncPgConnection {
         //
         // We apply this workaround to prevent requiring all the diesel
         // serialization code to beeing async
-        let mut metadata_lookup = PgAsyncMetadataLookup::new();
+        let mut bind_collector_0 = RawBytesBindCollector::<diesel::pg::Pg>::new();
+        let collect_bind_result_0 = query.collect_binds(
```
weiznich, line 370: We want to have some comment here why these steps are necessary and how the approach works from a top level view. Just to make sure that future refactoring won't break it again.

weiznich (commented):
I finally found some time to go over the changes and note some thoughts. 

First of all I would like to thank you for working on this and finding this solution. That is really good work. 

I have notes some ideas how to side step the more expensive parts of this approach that I think might be worth to try out, even if the introduce more complexity.

weiznich (approved):

weiznich:
I've pushed a bunch of cleanups + optimizations. This should be good to go now.

PR #160, opened 2024-06-19
sync wrapper: Clear mutex poison and provide underlying mutex guar

momobel:
## Issue

Sqlite can [return busy](https://sqlite.org/rescode.html#busy) especially when multiple transactions are running in concurrently. In async context with the Sync Wrapper, this will cause one client to get error (1) and panic within diesel causing the mutex to be poisoned and all subsequent clients panicking as a a result.

## Reason

A [busy timeout](https://sqlite.org/c3ref/busy_timeout.html) can be set and sqlite will pause waiting clients. This is set by the application but theoretically panics can still happen even with values in the order of seconds. Therefore an application could have a critical failure.

## Resolution

This PR tries to recover from such situations. It includes a commit allow reproducing the issue that could be dropped if the PR is validated.

Would you have any other solutions to this issue? Or do you think this is the most we can do at the moment?

## Notes

[1]
A [panic is triggered](https://github.com/diesel-rs/diesel/blob/master/diesel/src/sqlite/connection/statement_iterator.rs#L165) because the statement iterator is `NotStarted` and doesn't contain the statement (that was consumed below).
Because an [error is returned during the step](https://github.com/diesel-rs/diesel/blob/master/diesel/src/sqlite/connection/statement_iterator.rs#L104).
I am not sure we can recover except by retrying until getting a non-busy result. This is still not ideal because it would be blocking or with arbitrary limits.

### DIFF
--- a/examples/sync-wrapper/src/main.rs
+++ b/examples/sync-wrapper/src/main.rs
@@ -4,6 +4,7 @@ use diesel_async::async_connection_wrapper::AsyncConnectionWrapper;
 use diesel_async::sync_connection_wrapper::SyncConnectionWrapper;
 use diesel_async::{AsyncConnection, RunQueryDsl, SimpleAsyncConnection};
 use diesel_migrations::{embed_migrations, EmbeddedMigrations, MigrationHarness};
+use futures_util::FutureExt;
 
 // ordinary diesel model setup
 
@@ -15,7 +16,7 @@ table! {
 }
 
 #[allow(dead_code)]
-#[derive(Debug, Queryable, Selectable)]
+#[derive(Debug, Queryable, QueryableByName, Selectable)]
 #[diesel(table_name = users)]
 struct User {
     id: i32,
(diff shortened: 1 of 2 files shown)


weiznich:
Thanks for opening this PR. I'm currently away for holidays. I will have a look at this PR as after I'm back.

examples/sync-wrapper/src/main.rs
```diff
@@ -86,5 +119,32 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         .await?;
     println!("{data:?}");
 
+    // let changed = transaction(&mut sync_wrapper, "iLuke", "JustLuke").await?;
+    // println!("Changed {changed:?}");
```
weiznich, line 123:
```suggestion
```

src/sync_connection_wrapper.rs
```diff
@@ -206,9 +206,11 @@ impl<C> SyncConnectionWrapper<C> {
...
         let inner = self.inner.clone();
         tokio::task::spawn_blocking(move || {
-            let mut inner = inner
-                .lock()
-                .expect("Mutex is poisoned, a thread must have panicked holding it.");
+            let mut inner = inner.lock().unwrap_or_else(|poison| {
+                // try to be resilient by providing the guard
+                inner.clear_poison();
+                poison.into_inner()
+            });
```
weiznich, line 213: I just wonder whether it would be simpler to just always call `.clear_poison()` before lock here?

weiznich (approved):
Thanks for submitting this PR. This change looks conceptually good, but I would like to see the "test" moved to the actual tests.

weiznich (approved):

momobel:
Sorry, just came back to this topic and saw that you made comments. Thanks for handling them and merging!

PR #163, opened 2024-07-04
Update Rustls example and add to README for clarification

Brobb954:
As of December 2023 there was a change in the rustls crate that no long requires the collection of certs in a vec, I removed that and change the param for the next function accordingly. Also included a short blurb in the readme to specifically mention the requirement for tls if using an sslmode=require database


weiznich (approved):
Thanks :heart: 

I've pushed some more changes to use the latest rustls version and fix the compiler errors.

PR #174, opened 2024-07-24
opt out of `Send` for wasm32

insipx:
I'm working on adding a Wasm SQLite Backend to Diesel with the newly released [`wa-sqlite`](https://github.com/rhashimoto/wa-sqlite). 

Lots of functions for wa-sqlite are async and cannot be made sync in a wasm environment through blocking (b/c web). This means i've turned to this crate to try and continue to get it working, but this crate requires `Send` for it's Connection/diesel traits. This PR relaxes that condition for wasm32 environment.

### DIFF
--- a/src/pooled_connection/mod.rs
+++ b/src/pooled_connection/mod.rs
@@ -293,6 +293,7 @@ where
     }
 }
 
+#[allow(dead_code)]
 #[derive(diesel::query_builder::QueryId)]
 struct CheckConnectionQuery;
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -115,7 +115,8 @@ pub use self::transaction_manager::{AnsiTransactionManager, TransactionManager};
 /// Perform simple operations on a backend.
 ///
 /// You should likely use [`AsyncConnection`] instead.
-#[async_trait::async_trait]
+#[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
 pub trait SimpleAsyncConnection {
     /// Execute multiple SQL statements within the same string.
     ///
@@ -129,7 +130,8 @@ pub trait SimpleAsyncConnection {
 /// This trait represents a n async database connection. It can be used to query the database through
 /// the query dsl provided by diesel, custom extensions or raw sql queries. It essentially mirrors
 /// the sync diesel [`Connection`](diesel::connection::Connection) implementation
-#[async_trait::async_trait]
+#[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
 pub trait AsyncConnection: SimpleAsyncConnection + Sized + Send {
     /// The future returned by `AsyncConnection::execute`
     type ExecuteFuture<'conn, 'query>: Future<Output = QueryResult<usize>> + Send;
--- a/src/run_query_dsl/mod.rs
+++ b/src/run_query_dsl/mod.rs
@@ -695,7 +695,8 @@ impl<T, Conn> RunQueryDsl<Conn> for T {}
 /// #     Ok(())
 /// # }
 /// ```
-#[async_trait::async_trait]
+#[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
 pub trait SaveChangesDsl<Conn> {
     /// See the trait documentation
     async fn save_changes<T>(self, connection: &mut Conn) -> QueryResult<T>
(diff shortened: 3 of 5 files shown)


src/pooled_connection/mod.rs
```diff
@@ -293,6 +293,7 @@ where
     }
 }
 
+#[allow(dead_code)]
```
insipx, line 296: Haven't touched this part of the code but CI for rust beta doesn't pass w/o this. LMK if you want me to revert

weiznich:
Thanks for opening this PR. Can you explain why you need to relax this bound on a technical level. So what exactly does not work if that bound is there?  I can understand that it might be feel duplicated on a plain `wasm32` target but:

* I would really like to not have different behavior on different platforms if somewhat possible
* Even the `wasm32-unknown-unknow` target now allows some sort of threading via web-workers, although that requires setting a few additional target feature flags while compiling.

insipx:
@weiznich `wasm-bindgen` futures aren't `Send`, but async-trait requires `Send` by default. So, for instance if I try returning the future created with a `wasm-bindgen` JS Export, it will return a future that's `!Send` and won't satisfy the requirements for `Connection`. 

I understand the hesitation of having different behavior on different platforms, and I'd be OK leaving this as a duplicated/temp workaround until the `diesel-wasm-sqlite` backend becomes more full-featured. However, it would be nice to eventually have something like this. I think it's also common practice in other libraries, if that makes it any better.

weiznich:
Thanks for providing that context. That's at least helpful to understand why you made this change. For me that raises a more fundamental question: Why do you believe that the `AsyncConnection` trait is the better solution there instead of using the `Connection` trait from diesel itself? I mean if you don't have threads at all there is no way to do something like `spawn_blocking` + sqlite only offers a blocking API at all. Given that I personally do see no advantage of trying to use an async interface there at all.

> Maybe it would be worth putting a warning in the readme that the wasm target is experimental?

Platform support for diesel essentially boils down to: Officially supported is what we test in CI (in terms of target, platform, database system). Anything else is not supported. It might work, but we don't give any guarantee about it. I think that is quite similar to what most other projects do.

insipx:
> Thanks for providing that context. That's at least helpful to understand why you made this change. For me that raises a more fundamental question: Why do you believe that the AsyncConnection trait is the better solution there instead of using the Connection trait from diesel itself? I mean if you don't have threads at all there is no way to do something like spawn_blocking + sqlite only offers a blocking API at all. Given that I personally do see no advantage of trying to use an async interface there at all.



Yeah this is actually a very interesting problem. Sqlite does have only blocking calls, but the `wa-sqlite` has a largely Promise-based API that translates to Rust futures with Wasm Bindgen. For instance, consider the [exec](https://rhashimoto.github.io/wa-sqlite/docs/interfaces/SQLiteAPI.html#exec.exec-1) function of wa-sqlite, or [open_v2](https://rhashimoto.github.io/wa-sqlite/docs/interfaces/SQLiteAPI.html#open_v2.open_v2-1) which both return Promises. Even the 'sync' wasm build uses Promise-based functions.

As for why that is, i'm not entirely sure. There's probably an answer somewhere in [wa-sqlite discussions](https://github.com/rhashimoto/wa-sqlite/discussions)

According to [this discussion](https://github.com/rhashimoto/wa-sqlite/discussions/63#discussion-4541534), it may be something to do with browser APIS like OPFS

weiznich:
So just to summarize everything: It's basically required as `wa-sqlite` bindings generate futures that are not `Send`, right?

insipx:
> So just to summarize everything: It's basically required as wa-sqlite bindings generate futures that are not Send, right?

Yes :) 

There is a recent discussion that popped up on the topic of a sync wa-sqlite [here](https://github.com/rhashimoto/wa-sqlite/discussions/200). It looks out of scope for the current `wa-sqlite` creators, but it would definitely be interesting to watch the progress in the future

weiznich:
I had some time thinking about this again. As previously mentioned I do not like the idea to couple this on the target. In the end it's something that depends on the underlying connection implementation, not on the target. 

Given that I would like to propose a different solution: We move away from using `#[async_trait]` and instead have an explicit associated type per function returning a future. That associated type would then be only require `Future`, but it would allow implementations to opt into the `+ Send` by using `Box<dyn Future + Send>`. I think for `AsyncConnection` that should be relatively straight forward. For traits building on top of these it will be a bit more complicated. I expect that this will either require to specify complex contaminator types there or having a custom future type so that the compiler can automatically infer if the returned type is `Send` or not. Being able to just use `impl Future` there would be great, but it seams like you cannot use it for associated types yet. For trait functions it doesn't "leak" the `Send` bound as it's the case for free standing functions, which in turn makes it not usable there.

insipx:
That makes sense, and overall a more robust solution. Even with this PR, I had to end up replicating a bunch of the trait definitions to fit them into WASM. 

Since I don't have the bandwidth to work on this myself right now, can close this and I opened https://github.com/weiznich/diesel_async/issues/185 instead

PR #177, opened 2024-08-06
Add Clone for SyncConnectionWrapper

pickfire:
Clone is needed for some async stuff like teloxide.

See discussion in https://github.com/teloxide/teloxide/discussions/1124


pickfire:
Thanks, that seemed to be able to solve the issue.

PR #188, opened 2024-09-17
[features] add `pool` crate feature

fakeshadow:
Related discussion: #187


weiznich (approved):
Looks good, thanks for opening this PR :+1:

Just to confirm: You are mostly interested in having access to the `PoolableConnection` trait for your third party connection type, right?

fakeshadow:
> Looks good, thanks for opening this PR 👍
> 
> Just to confirm: You are mostly interested in having access to the `PoolableConnection` trait for your third party connection type, right?

Thanks for the quick review and merge. 
Yes the `PoolableConnection` trait is essential for both my connection type and my pooling type. With it my connection type can use pooling feature from `diesel-async` like `bb8` and `deadpool`. And at the same time my pooling type can use connection type from `diesel-async` as `AsyncPgConnection` and `AsyncMySqlConnection`.
Other public types and trait from `pooled_connection` are good to have for keeping a consistent api setup but they are not must keeping as public.

PR #193, opened 2024-10-28
CI: Run `async-connection-wrapper` feature tests too

Turbo87:
`tests/sync_wrapper.rs` is gated by `#[cfg(feature = "async-connection-wrapper")]`. Since the `async-connection-wrapper` feature wasn't explicitly enabled, we were not actually running these tests on CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -202,7 +202,7 @@ jobs:
         run: cargo +${{ matrix.rust }} version
 
       - name: Test diesel_async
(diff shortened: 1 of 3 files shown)


Turbo87:
looks like the `sqlite` variant of one of the tests was unhappy. I've added another commit that should fix it.

Turbo87:
looks like this is finally green now too 😅

weiznich (approved):
Thanks, I will  try to cut a patch release with the deref changes soon.

PR #199, opened 2024-11-21
Derive Debug for SyncConnectionWrapper

tibbe:
This is needed when trying to use [tower-sessions](https://crates.io/crates/tower-sessions) with this crate. `tower-sessions` requires `Debug` for the provided backend store. In my particular case the store looks like:

```rust
type SqlitePool = deadpool::Pool<SyncConnectionWrapper<sqlite::SqliteConnection>>;

#[derive(Clone, Debug)]
pub struct AsyncSqliteStore {
    pool: SqlitePool,
}
```

The error without this `Debug` impl is:

```
`diesel_async::sync_connection_wrapper::SyncConnectionWrapper<diesel::SqliteConnection>` doesn't implement `std::fmt::Debug`
the trait `std::fmt::Debug` is not implemented for `diesel_async::sync_connection_wrapper::SyncConnectionWrapper<diesel::SqliteConnection>`, which is required by `deadpool::managed::Pool<diesel_async::pooled_connection::AsyncDieselConnectionManager<diesel_async::sync_connection_wrapper::SyncConnectionWrapper<diesel::SqliteConnection>>>: std::fmt::Debug`
```

### DIFF
--- a/src/sync_connection_wrapper/mod.rs
+++ b/src/sync_connection_wrapper/mod.rs
@@ -73,6 +73,7 @@ fn from_tokio_join_error(join_error: JoinError) -> diesel::result::Error {
 /// #    some_async_fn().await;
 /// # }
 /// ```
+#[derive(Debug)]
 pub struct SyncConnectionWrapper<C> {
     inner: Arc<Mutex<C>>,
 }


tibbe:
The particular trait I need to implement that required `Debug` is https://docs.rs/tower-sessions/0.13.0/tower_sessions/trait.SessionStore.html.

tibbe:
The MacOS build failures seem unrelated.

src/sync_connection_wrapper/mod.rs
```diff
@@ -73,6 +73,7 @@ fn from_tokio_join_error(join_error: JoinError) -> diesel::result::Error {
 /// #    some_async_fn().await;
 /// # }
 /// ```
+#[derive(Debug)]
 pub struct SyncConnectionWrapper<C> {
     inner: Arc<Mutex<C>>,
```
weiznich, line 78: Thanks for submitting this PR.
Can you explain why this would be actual helpful for your use-case as `#[derive(Debug)]` only implements `Debug` if all generic parameters (in this case `C`) implements `Debug` on their own. I think all of the possible types for `C` (especially the `SqliteConnection` type) do not implement `Debug` on their own, so this won't actually be used at all.

tibbe: It's a first step. I think need to work on e.g. `SqliteConnection`.

weiznich: As I had that discussion already a few times: The connection types do not implement `Debug` because there is nothing you can reasonably show there. They only hold a few opaque pointers to c objects. We also do not want to store the database URL in their as that could leak confidential information (login credentials, etc).

tibbe: Fair enough. I will close this PR then.

PR #204, opened 2024-12-16
chore: upgrade `bb8` to `v9.0`

sabify:
`bb8` `v9.0` dropped `async_trait` and bumped MSRV to `v1.75.0` which is already compatible with this library's MSRV `v1.78.0`.

### DIFF
--- a/src/mysql/row.rs
+++ b/src/mysql/row.rs
@@ -37,7 +37,11 @@ impl RowSealed for MysqlRow {}
 
 impl<'a> diesel::row::Row<'a, Mysql> for MysqlRow {
     type InnerPartialRow = Self;
-    type Field<'b> = MysqlField<'b> where Self: 'b, 'a: 'b;
+    type Field<'b>
+        = MysqlField<'b>
+    where
+        Self: 'b,
+        'a: 'b;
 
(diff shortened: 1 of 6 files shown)


src/mysql/row.rs
```diff
@@ -1,9 +1,15 @@
...
+    mysql::{
+        data_types::{MysqlTime, MysqlTimestampType},
+        Mysql, MysqlType, MysqlValue,
+    },
+    row::{PartialRow, RowIndex, RowSealed},
+};
+use mysql_async::{
+    consts::{ColumnFlags, ColumnType},
+    Column, Row, Value,
+};
```
weiznich, line 12: I would prefer not to reformat all imports. Could you roll back all import formatting related changes?

weiznich (changes requested):
Thanks for opening this PR. The actual change itself look good, although the formatting related changes (mostly the imports) shouldn't be in here.

src/mysql/row.rs
sabify: Sure, please checkout the latest changes.

weiznich (approved):
Looks good now, thanks for the fast fix

PR #208, opened 2025-01-15
[Update] examples to use rustls-platform-verifier

talves:
Replace `rustls-native-certs` with `rustls-platform-verifier` which will default to the correct certs use based on platform. This is recommended by the `rustls-native-certs` owner and is still being used within the `rustls-platform-verifier`.


weiznich (approved):
Looks good, thanks for submitting

Issue #142, opened 2024-03-13
RunQueryDsl overrides calls to vec.first() to RunQueryDsl::first()


cdxker:
## Setup

RunQueryDsl overrides calls to vec.first to RunQueryDsl::first() this happens to all possible calls to first() if the trait is imported.

### Versions

- **Rust:**: 1.75
- **Diesel:**: 2.1.4
- **Diesel_async:**: 0.4.1
- **Database:**: None
- **Operating System**: EndeavourOS

### Feature Flags

- **diesel:**: None
- **diesel_async:**: None

## Problem Description

The `diesel_async::RunQueryDsl` trait overrides the default behavior of any structs `first()` function to RunQueryDsl::first causing a nasty compile time error.

### What are you trying to accomplish?

 I wish to use the vec.first() operation instead of vec.get(0). (Mainly becuase clippy errors, but this is a pretty hard bug to find).

### Steps to reproduce

Here is a gist that shows the issue at hand.
https://gist.github.com/fastfists/9d0f1da80f825be39f9a4499aa8990a7

## Checklist

- [X] I have already looked over the [issue tracker](https://github.com/weiznich/diesel_async/issues) for similar possible closed issues.
- [X] This issue can be reproduced on Rust's stable channel. (Your issue will be
  closed if this is not the case)
- [X] This issue can be reproduced without requiring a third party crate

weiznich:
Thanks for filling this bug report. I've changed it to be tracked as feature request, as it's strictly speaking not a bug. We are happy to receive a PR that improves this as long as it does not break existing code.

fastfists:
I tried reading into it but did not quite understand, how does the trait work at the current moment? Does it try to infer any `diesel::QueryDsl` functions into `diesel_async::RunQueryDsl` functions?

weiznich:
The relevant implementation is here: https://github.com/weiznich/diesel_async/blob/1ef43d2a8a4080c1e9e5c34056f4d26cbb2a85cc/src/run_query_dsl/mod.rs#L626-L637

That one essentially says that this method is only available for types that implement `LimitDsl`. That shouldn't be the case for `Vec<T>`, because that trait is only implemented for diesel query DSL structs. So I'm not sure why rustc believes that this could potentially overlap with the `first` method from `Vec<()>`, might be even worth filling an issue for that at the rust repo? (Feel free to ping me from there)

fiadliel:
[...]

PR #214, opened 2025-02-05
RunQueryDsl: match same trait constraints as diesel

eakoli:
This resolves issue https://github.com/weiznich/diesel_async/issues/142 which stems from the async RunQueryDsl being implemented for ALL types.

In addition it adds some constraints that were missing from async LoadQuery when compared to diesel LoadQuery

Comparing the original diesel RunQueryDsl to the async one, the original only implements RunQueryDsl for types that implement Table and a few other explicitly tagged (like SqlQuery)

https://github.com/diesel-rs/diesel/blob/v2.2.7/diesel/src/query_dsl/mod.rs#L1785-L1791

the following shows how there is not an issue with diesel proper, only with diesel_async::RunQueryDsl

```rust
mod sub {

    pub(crate) struct A {
        pub(crate) items: Vec<String>,
    }
}

mod success_1 {
    fn test(value: super::sub::A) -> String {
        value.items.first().map(|s| s.into()).unwrap() // properly compiles and uses Vec::first
    }
}

mod success_2 {
    use diesel::prelude::*;
    fn test(value: super::sub::A) -> String {
        value.items.first().map(|s| s.into()).unwrap() // properly compiles and uses Vec::first
    }
}

mod failure {
    use diesel::prelude::*;
    use diesel_async::RunQueryDsl;

    fn test(value: super::sub::A) -> String {
        value.items.first().map(|s| s.into()).unwrap() // complains Table is not implemented by Vec<std::string::String>
    }
}
```

### DIFF
--- a/src/run_query_dsl/mod.rs
+++ b/src/run_query_dsl/mod.rs
@@ -71,7 +71,7 @@ pub mod methods {
     /// to call `load` from generic code.
     ///
     /// [`RunQueryDsl`]: super::RunQueryDsl
-    pub trait LoadQuery<'query, Conn: AsyncConnectionCore, U> {
+    pub trait LoadQuery<'query, Conn: AsyncConnectionCore, U>: RunQueryDsl<Conn> {
         /// The future returned by [`LoadQuery::internal_load`]
         type LoadFuture<'conn>: Future<Output = QueryResult<Self::Stream<'conn>>> + Send
         where
@@ -91,7 +91,7 @@ pub mod methods {
         Conn: AsyncConnectionCore<Backend = DB>,
         U: Send,
         DB: Backend + 'static,
-        T: AsQuery + Send + 'query,
+        T: AsQuery + RunQueryDsl<Conn> + Send + 'query,
         T::Query: QueryFragment<DB> + QueryId + Send + 'query,
         T::SqlType: CompatibleType<U, DB, SqlType = ST>,
         U: FromSqlRow<ST, DB> + Send + 'static,
@@ -589,7 +589,8 @@ pub trait RunQueryDsl<Conn>: Sized {
     }
 }
 
-impl<T, Conn> RunQueryDsl<Conn> for T {}
+// Use marker trait to implement RunQueryDsl
+impl<T, Conn> RunQueryDsl<Conn> for T where T: diesel::query_dsl::SupportRunQueryDsl {}
 
 /// Sugar for types which implement both `AsChangeset` and `Identifiable`
 ///
--- a/src/run_query_dsl/mod.rs
+++ b/src/run_query_dsl/mod.rs
@@ -71,7 +71,7 @@ pub mod methods {
     /// to call `load` from generic code.
     ///
     /// [`RunQueryDsl`]: super::RunQueryDsl
-    pub trait LoadQuery<'query, Conn: AsyncConnectionCore, U>: RunQueryDsl<Conn> {
+    pub trait LoadQuery<'query, Conn: AsyncConnectionCore, U> {
         /// The future returned by [`LoadQuery::internal_load`]
         type LoadFuture<'conn>: Future<Output = QueryResult<Self::Stream<'conn>>> + Send
         where
@@ -91,7 +91,7 @@ pub mod methods {
         Conn: AsyncConnectionCore<Backend = DB>,
         U: Send,
         DB: Backend + 'static,
-        T: AsQuery + RunQueryDsl<Conn> + Send + 'query,
+        T: AsQuery + Send + 'query,
         T::Query: QueryFragment<DB> + QueryId + Send + 'query,
         T::SqlType: CompatibleType<U, DB, SqlType = ST>,
         U: FromSqlRow<ST, DB> + Send + 'static,
@@ -583,14 +583,14 @@ pub trait RunQueryDsl<Conn>: Sized {
         U: Send + 'conn,
         Conn: AsyncConnectionCore,
         Self: diesel::query_dsl::methods::LimitDsl,
-        diesel::dsl::Limit<Self>: methods::LoadQuery<'query, Conn, U> + Send + 'query,
+        diesel::dsl::Limit<Self>:
+            methods::LoadQuery<'query, Conn, U> + RunQueryDsl<Conn> + Send + 'query,
     {
         diesel::query_dsl::methods::LimitDsl::limit(self, 1).get_result(conn)
     }
 }
 
-// Use marker trait to implement RunQueryDsl
-impl<T, Conn> RunQueryDsl<Conn> for T where T: diesel::query_dsl::SupportRunQueryDsl {}
+impl<T, Conn> RunQueryDsl<Conn> for T where T: diesel::query_dsl::RunQueryDslSupport {}
 
 /// Sugar for types which implement both `AsChangeset` and `Identifiable`
 ///
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -115,3 +115,6 @@ members = [
   "examples/postgres/run-pending-migrations-with-rustls",
   "examples/sync-wrapper",
 ]
+
+[patch.crates-io]
+diesel =  { git = "https://github.com/eakoli/diesel/", branch = "feat/run-query-dsl-marker-trait" }
--- a/tests/lib.rs
+++ b/tests/lib.rs
@@ -12,6 +12,7 @@ mod migrations;
 mod notifications;
 #[cfg(any(feature = "bb8", feature = "deadpool", feature = "mobc"))]
 mod pooling;
+mod run_query_dsl_impl_match;
 #[cfg(feature = "async-connection-wrapper")]
 mod sync_wrapper;
 mod transactions;
--- a/tests/run_query_dsl_impl_match.rs
+++ b/tests/run_query_dsl_impl_match.rs
@@ -1,0 +1,148 @@
+use super::{connection, users, User};
+use diesel::prelude::*;
+use diesel_async::RunQueryDsl;
+
+#[derive(Debug, QueryableByName, PartialEq)]
+struct CountType {
+    #[diesel(sql_type = diesel::sql_types::BigInt)]
+    pub count: i64,
+}
+
+#[tokio::test]
+async fn test_boxed_sql() {
+    use diesel::query_builder::BoxedSqlQuery;
+
+    let mut con = connection().await;
+
+    let boxed: BoxedSqlQuery<_, _> =
+        diesel::sql_query("select count(*) as count from users").into_boxed();
+
+    let result = boxed.get_result::<CountType>(&mut con).await.unwrap();
+
+    assert_eq!(result.count, 0);
+}
+
+#[tokio::test]
+async fn test_select_statement() {
+    use diesel::query_builder::SelectStatement;
+
+    let mut con = connection().await;
+
+    let select: SelectStatement<_, _, _, _, _, _, _> = users::table.select(User::as_select());
+
+    let result = select.get_results(&mut con).await.unwrap();
+
+    assert_eq!(result.len(), 0);
+}
+
+#[tokio::test]
+async fn test_sql_query() {
+    use diesel::query_builder::SqlQuery;
+
+    let mut con = connection().await;
+
+    let sql: SqlQuery<_> = diesel::sql_query("select count(*) as count from users");
+
+    let result: CountType = sql.get_result(&mut con).await.unwrap();
+
+    assert_eq!(result.count, 0);
+}
+
+#[tokio::test]
+async fn test_unchecked_bind() {
+    use diesel::expression::UncheckedBind;
+    use diesel::sql_types::{BigInt, Text};
+
+    let mut con = connection().await;
+
+    let unchecked: UncheckedBind<_, _> =
+        diesel::dsl::sql::<BigInt>("SELECT count(id) FROM users WHERE name = ")
+            .bind::<Text, _>("Bob");
+
+    let result = unchecked.get_result(&mut con).await;
+
+    assert_eq!(Ok(0), result);
+}
+
+#[tokio::test]
+async fn test_alias() {
+    use diesel::query_source::Alias;
+
+    let mut con = connection().await;
+
+    let aliased: Alias<_> = diesel::alias!(users as other);
+
+    let result = aliased.get_results::<User>(&mut con).await.unwrap();
+
+    assert_eq!(result.len(), 0);
+}
+
+#[tokio::test]
+async fn test_boxed_select() {
+    use diesel::query_builder::BoxedSelectStatement;
+
+    let mut con = connection().await;
+
+    let select: BoxedSelectStatement<_, _, _, _> =
+        users::table.select(User::as_select()).into_boxed();
+
+    let result = select.get_results(&mut con).await.unwrap();
+
+    assert_eq!(result.len(), 0);
+}
+
+#[tokio::test]
+async fn test_sql_literal() {
+    use diesel::expression::SqlLiteral;
+    use diesel::sql_types::Integer;
+
+    let mut con = connection().await;
+
+    let literal: SqlLiteral<_> = diesel::dsl::sql::<Integer>("SELECT 6");
+
+    let result = literal.get_result(&mut con).await;
+
+    assert_eq!(Ok(6), result);
+}
+
+#[tokio::test]
+async fn test_delete() {
+    use diesel::query_builder::DeleteStatement;
+
+    let mut con = connection().await;
+
+    let delete: DeleteStatement<_, _, _> = diesel::delete(users::table);
+
+    let result = delete.execute(&mut con).await;
+
+    assert_eq!(Ok(0), result);
+}
+
+#[tokio::test]
+async fn test_insert_statement() {
+    use diesel::query_builder::InsertStatement;
+
+    let mut con = connection().await;
+
+    let inserted_names: InsertStatement<_, _, _, _> =
+        diesel::insert_into(users::table).values(users::name.eq("Timmy"));
+
+    let result = inserted_names.execute(&mut con).await;
+
+    assert_eq!(Ok(1), result);
+}
+
+#[tokio::test]
+async fn test_update_statement() {
+    use diesel::query_builder::UpdateStatement;
+
+    let mut con = connection().await;
+
+    let update: UpdateStatement<_, _, _, _> = diesel::update(users::table)
+        .set(users::name.eq("Jim"))
+        .filter(users::name.eq("Sean"));
+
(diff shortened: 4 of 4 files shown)


src/run_query_dsl/mod.rs
```diff
@@ -641,7 +641,9 @@ pub trait RunQueryDsl<Conn>: Sized {
     }
 }
 
-impl<T, Conn> RunQueryDsl<Conn> for T {}
+// Note: Match the same types that diesel::RunQueryDsl applies to this
+// seems safe currently, as the trait imposes no restrictions based on Conn, only on T.
+impl<T, Conn> RunQueryDsl<Conn> for T where T: diesel::RunQueryDsl<Conn> {}
```
weiznich, line 646: I fear this assumption is not correct. See for example this impl: https://docs.diesel.rs/master/diesel/prelude/trait.RunQueryDsl.html#impl-RunQueryDsl%3CConn%3E-for-BoxedSqlQuery%3C'_,+%3CConn+as+Connection%3E::Backend,+Query%3E

For the linked impl these trait bounds would imply that `Conn` implements `AsyncConnection` and `Connection` at the same time, which is generally not fulfilled by any type.

Also even if we get it so that diesel doesn't have this kind of constraints internally that wouldn't help with third party crates which totally could write connection specific impls.

weiznich (commented):
Thanks for submitting this PR. I fear this approach won't work for the reasons outlined below.

src/run_query_dsl/mod.rs
eakoli: Im not sure I follow what the reason is, are you saying that the where ends up being too restrictive?

> For the linked impl these trait bounds would imply that Conn implements AsyncConnection and Connection at the same time, which is generally not fulfilled by any type.

Where does it imply that Conn implements anything? It only makes assertion with respect to T, not to Conn.

>Also even if we get it so that diesel doesn't have this kind of constraints internally that wouldn't help with third party crates which totally could write connection specific impls.

What internal constraint are you referring to? And connection specific impl of what?


What about something a little more explicit?


```suggest
    impl <T, Conn: Connection, AsyncConn> RunQueryDsl<AsyncConn> for T where T: RunQueryDsl<Conn> {}
```

eakoli: Ok, I see what you mean with the conn :Connection, and unmatched type.

What about `T where T: AsQuery`, as [this comment](https://github.com/diesel-rs/diesel/blob/v2.2.7/diesel/src/query_dsl/mod.rs#L1785-L1791) implies it really should be ?

I pushed up a changes to match on AsQuery, as well as tests to verify all the current types that work for RunQueryDsl<conn : Connection> also work for diesel_async::RunQueryDsl<conn: AsyncConnection>

weiznich: Sorry for taking that long to answer, but I really did not have the time to actually check whether the comment there is still relevant. 
I did now have some time and checked with the following example:

```rust
async fn test(conn: &mut AsyncPgConnection) {
    users::table
        .filter(posts::id.eq(42))
        .load::<(i32,)>(conn)
        .await;
}
```

without your patch that generates the following error:

```
error[E0271]: type mismatch resolving `<table as AppearsInFromClause<table>>::Count == Once`
  --> src/main.rs:11:10
   |
11 |         .load::<(i32,)>(conn)
   |          ^^^^ expected `Once`, found `Never`
   |
note: required for `posts::columns::id` to implement `AppearsOnTable<users::table>`
  --> src/main.rs:6:25
   |
6  | diesel::table! { posts{ id -> Integer}}
   |                         ^^
   = note: associated types for the current `impl` cannot be restricted in `where` clauses
   = note: 2 redundant requirements hidden
   = note: required for `Grouped<Eq<id, Bound<Integer, i32>>>` to implement `AppearsOnTable<users::table>`
   = note: required for `WhereClause<Grouped<Eq<id, Bound<Integer, i32>>>>` to implement `query_builder::where_clause::ValidWhereClause<FromClause<users::table>>`
   = note: required for `SelectStatement<FromClause<table>, DefaultSelectClause<...>, ..., ...>` to implement `Query`
   = note: required for `SelectStatement<FromClause<table>, DefaultSelectClause<...>, ..., ...>` to implement `diesel_async::methods::LoadQuery<'_, _, (i32,)>`
   = note: the full name for the type has been written to '/tmp/diesel_test/target/debug/deps/diesel_test-acb1381933ce240a.long-type-13903671074851835938.txt'
   = note: consider using `--verbose` to print the full type name to the console

error[E0277]: the trait bound `users::table: TableNotEqual<posts::table>` is not satisfied
  --> src/main.rs:11:10
   |
11 |         .load::<(i32,)>(conn)
   |          ^^^^ the trait `TableNotEqual<posts::table>` is not implemented for `users::table`
   |
   = note: double check that `posts::table` and `users::table` appear in the same `allow_tables_to_appear_in_same_query!` 
           call if both are tables
   = help: the following other types implement trait `TableNotEqual<T>`:
             `Only<diesel::pg::metadata_lookup::pg_namespace::table>` implements `TableNotEqual<diesel::pg::metadata_lookup::pg_type::table>`
             `Only<diesel::pg::metadata_lookup::pg_type::table>` implements `TableNotEqual<diesel::pg::metadata_lookup::pg_namespace::table>`
             `Tablesample<diesel::pg::metadata_lookup::pg_namespace::table, TSM>` implements `TableNotEqual<diesel::pg::metadata_lookup::pg_type::table>`
             `Tablesample<diesel::pg::metadata_lookup::pg_type::table, TSM>` implements `TableNotEqual<diesel::pg::metadata_lookup::pg_namespace::table>`
             `diesel::pg::metadata_lookup::pg_namespace::table` implements `TableNotEqual<Only<diesel::pg::metadata_lookup::pg_type::table>>`
             `diesel::pg::metadata_lookup::pg_namespace::table` implements `TableNotEqual<Tablesample<diesel::pg::metadata_lookup::pg_type::table, TSM>>`
             `diesel::pg::metadata_lookup::pg_namespace::table` implements `TableNotEqual<diesel::pg::metadata_lookup::pg_type::table>`
             `diesel::pg::metadata_lookup::pg_type::table` implements `TableNotEqual<Only<diesel::pg::metadata_lookup::pg_namespace::table>>`
           and 2 others
   = note: required for `users::table` to implement `AppearsInFromClause<posts::table>`
note: required for `posts::columns::id` to implement `AppearsOnTable<users::table>`
  --> src/main.rs:6:25
   |
6  | diesel::table! { posts{ id -> Integer}}
   |                         ^^
   = note: 2 redundant requirements hidden
   = note: required for `Grouped<Eq<id, Bound<Integer, i32>>>` to implement `AppearsOnTable<users::table>`
   = note: required for `WhereClause<Grouped<Eq<id, Bound<Integer, i32>>>>` to implement `query_builder::where_clause::ValidWhereClause<FromClause<users::table>>`
   = note: required for `SelectStatement<FromClause<table>, DefaultSelectClause<...>, ..., ...>` to implement `Query`
   = note: required for `SelectStatement<FromClause<table>, DefaultSelectClause<...>, ..., ...>` to implement `diesel_async::methods::LoadQuery<'_, _, (i32,)>`
   = note: the full name for the type has been written to '/tmp/diesel_test/target/debug/deps/diesel_test-acb1381933ce240a.long-type-13903671074851835938.txt'
   = note: consider using `--verbose` to print the full type name to the console
```

with your patch this generates that error:

```
error[E0599]: the method `load` exists for struct `SelectStatement<FromClause<table>, DefaultSelectClause<...>, ..., ...>`, but its trait bounds were not satisfied
  --> src/main.rs:11:10
   |
9  | /     users::table
10 | |         .filter(posts::id.eq(42))
11 | |         .load::<(i32,)>(conn)
   | |         -^^^^ method cannot be called due to unsatisfied trait bounds
   | |_________|
   |
   |
  ::: /home/weiznich/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/diesel-2.2.4/src/query_builder/select_statement/mod.rs:48:1
   |
48 | / #[diesel_derives::__diesel_public_if(
49 | |     feature = "i-implement-a-third-party-backend-and-opt-into-breaking-changes",
50 | |     public_fields(
51 | |         select,
...  |
61 | | )]
   | |__- doesn't satisfy `_: AsQuery` or `_: RunQueryDsl<_>`
   |
   = note: the full type name has been written to '/tmp/diesel_test/target/debug/deps/diesel_test-7a841758257f912e.long-type-11988839005461461334.txt'
   = note: consider using `--verbose` to print the full type name to the console
   = note: the following trait bounds were not satisfied:
           `SelectStatement<FromClause<users::table>, query_builder::select_clause::DefaultSelectClause<FromClause<users::table>>, query_builder::distinct_clause::NoDistinctClause, query_builder::where_clause::WhereClause<diesel::expression::grouped::Grouped<diesel::expression::operators::Eq<posts::columns::id, diesel::expression::bound::Bound<diesel::sql_types::Integer, i32>>>>>: AsQuery`
           which is required by `SelectStatement<FromClause<users::table>, query_builder::select_clause::DefaultSelectClause<FromClause<users::table>>, query_builder::distinct_clause::NoDistinctClause, query_builder::where_clause::WhereClause<diesel::expression::grouped::Grouped<diesel::expression::operators::Eq<posts::columns::id, diesel::expression::bound::Bound<diesel::sql_types::Integer, i32>>>>>: diesel_async::RunQueryDsl<_>`
           `&SelectStatement<FromClause<users::table>, query_builder::select_clause::DefaultSelectClause<FromClause<users::table>>, query_builder::distinct_clause::NoDistinctClause, query_builder::where_clause::WhereClause<diesel::expression::grouped::Grouped<diesel::expression::operators::Eq<posts::columns::id, diesel::expression::bound::Bound<diesel::sql_types::Integer, i32>>>>>: AsQuery`
           which is required by `&SelectStatement<FromClause<users::table>, query_builder::select_clause::DefaultSelectClause<FromClause<users::table>>, query_builder::distinct_clause::NoDistinctClause, query_builder::where_clause::WhereClause<diesel::expression::grouped::Grouped<diesel::expression::operators::Eq<posts::columns::id, diesel::expression::bound::Bound<diesel::sql_types::Integer, i32>>>>>: diesel_async::RunQueryDsl<_>`
           `&mut SelectStatement<FromClause<users::table>, query_builder::select_clause::DefaultSelectClause<FromClause<users::table>>, query_builder::distinct_clause::NoDistinctClause, query_builder::where_clause::WhereClause<diesel::expression::grouped::Grouped<diesel::expression::operators::Eq<posts::columns::id, diesel::expression::bound::Bound<diesel::sql_types::Integer, i32>>>>>: AsQuery`
           which is required by `&mut SelectStatement<FromClause<users::table>, query_builder::select_clause::DefaultSelectClause<FromClause<users::table>>, query_builder::distinct_clause::NoDistinctClause, query_builder::where_clause::WhereClause<diesel::expression::grouped::Grouped<diesel::expression::operators::Eq<posts::columns::id, diesel::expression::bound::Bound<diesel::sql_types::Integer, i32>>>>>: diesel_async::RunQueryDsl<_>`
   = help: items from traits can only be used if the trait is in scope
```

So it seems like the comment over there in the diesel code is still relevant today. I'm sorry to say that, but in my opinion that's a too bad regression of the error messages emitted in such failure cases to be acceptable.

urkle:
@weiznich is there anyway we can get something in place to reduce this global injection?  As this makes it very challenging to use diesel-async without having to always use the alternate trait call syntax. As I have other crates which implement same-named methods (e.g. execute etc.) that start conflicting all over.

```rust
diesel_async::RunQueryDsl::first::<Model>(query, &mut db);
```

eakoli:
@weiznich Ive been experimenting with two alternative measures to remove the global injection.

A) if async was moved into diesel proper (as opposed to being an extra create) then it could properly apply the same exact impl clauses as the sync RunQueryDsl.

B) (possibly less intrusive) introduce a marker trait with no bounds to indicate what should implement RunQueryDsl, similar to how its currently matched on `T: Table`
```
  // Marker trait for opting in to RunQueryDsl support
  trait SupportRunQueryDsl {}

  // items that currently implement RunQueryDsl explicitly
  // -- query_dsl/mod.rs
  - impl<T, Conn> RunQueryDsl<Conn> for T: Table {};
  + impl<T> SupportRunQueryDsl for T: Table {};

  // -- query_dsl/positional_order_dsl.rs
  - impl<Source, Expr, Conn> RunQueryDsl<Conn> for PositionalOrderClause<Source, Expr> {}
  + impl<Source, Expr> SupportRunQueryDsl for PositionalOrderClause<Source, Expr> {}

  // -- expression/sql_literal.rs
  - impl<ST, T, Conn> RunQueryDsl<Conn> for SqlLiteral<ST,T> {}
  + impl<ST, T> SupportRunQueryDsl for SqlLiteral<ST,T> {}
  ...
  - impl<Query, Value, Conn> RunQueryDsl<Conn> for UncheckedBind<Query,Value> {} 
  + impl<Query, Value> SupportRunQueryDsl for UncheckedBind<Query,Value> {}

  // -- query_builder/sql_query.rs
  - impl<Inner, Conn> RunQueryDsl<Conn> for diesel::query_builder::SqlQuery<Inner> {}
  + impl<Inner> SupportRunQueryDsl for diesel::query_builder::SqlQuery<Inner> {}
  ...
  - impl<Conn, Query, Value> RunQueryDsl<Conn> for diesel::query_builder::UncheckedBind<Query,Value> {}
  + impl<Query, Value> SupportRunQueryDsl for diesel::query_builder::UncheckedBind<Query,Value> {}
  ...
  - impl<Conn: Connection, Query> RunQueryDsl<Conn> for diesel::query_builder::BoxedSqlQuery<'_, Conn::Backend, Query> {}
  + impl<Conn: Connection, Query> SupportRunQueryDsl for diesel::query_builder::BoxedSqlQuery<'_, Conn::Backend, Query> {}

 // -- query_builder/combination_clause.rs
 - impl<Combinator, Rule, Source, Rhs, Conn> RunQueryDsl<Conn> for CombinationClause<Combinator, Rule, Source, Rhs> {}
 + impl<Combinator, Rule, Source, Rhs> SupportRunQueryDsl for CombinationClause<Combinator, Rule, Source, Rhs> {}

 // -- query_builder/update_statement/mod.rs
 - impl<T: QuerySource, U, V, Ret, Conn> RunQueryDsl<Conn> for UpdateStatement<T, U, V, Ret> {}
 + impl<T: QuerySource, U, V, Ret> SupportRunQueryDsl for UpdateStatement<T, U, V, Ret> {}

 // -- query_builder/select_statement/boxed.rs
 - impl<ST, QS, DB, Conn, GB> RunQueryDsl<Conn> for BoxedSelectStatement<'_, ST, QS, DB, GB> {}
 + impl<ST, QS, DB, GB> SupportRunQueryDsl for BoxedSelectStatement<'_, ST, QS, DB, GB> {}

 // -- query_builder/select_statement/dsl_impls.rs
 - impl<F, S, D, W, O, LOf, G, H, LC, Conn> RunQueryDsl<Conn> for SelectStatement<F, S, D, W, O, LOf, G, H, LC> {}
 + impl<F, S, D, W, O, LOf, G, H, LC> SupportRunQueryDsl for SelectStatement<F, S, D, W, O, LOf, G, H, LC> {}

 // -- query_builder/insert_statement/mod.rs
 - impl<T: QuerySource, U, Op, Ret, Conn> RunQueryDsl<Conn> for SelectStatement<T, U, Op, Ret> {}
 + impl<T: QuerySource, U, Op, Ret> SupportRunQueryDsl for SelectStatement<T, U, Op, Ret> {}

 // -- query_builder/delete_statement/mod.rs
 - impl<T, U, Ret, Conn> RunQueryDsl<Conn> for DeleteStatement<T, U, Ret> where T: QuerySource {}
 + impl<T, U, Ret> SupportRunQueryDsl for DeleteStatement<T, U, Ret> where T: QuerySource {}

 // -- query_source/aliasing/dsl_impls.rs
 -  impl<S: AliasSource, Conn> RunQueryDsl<Conn> for Alias<S> {}
 +  impl<S: AliasSource> SupportRunQueryDsl for Alias<S> {}
 
  ....

  
  // then to implement RunQueryDsl on all things that should be is simply
  impl <T, Conn> RunQueryDsl<Conn> for T where T: SupportRunQueryDsl {}

  // which then lets diesel_async simply do this
  impl <T, Conn> diesel_async::RunQueryDsl<Conn> for T where T: SupportRunQueryDsl {}
```

C) Another option (although possibly a more breaking API change) would be to remove the Conn bounds from the RunQueryDsl trait (I don't think its really needed anyway, as all the methods take a conn and apply the actual bounds to it)

that way this solution would work, as there are no unconstrained parameters
```
 Note: Match the same types that diesel::RunQueryDsl applies to this
// seems safe currently, as the trait imposes no restrictions based on Conn, only on T.
impl<T> diesel_async::RunQueryDsl for T where T: diesel::RunQueryDsl {}
```


If any of these seem feasible, I can start working up the changes.

eakoli:
@weiznich RE: OptionB

I guess the marker would not need to replace the existing, but could be done in-addition to. That way nothing really changes in the public API, other than there being an additional trait that can be used to match on for diesel_async, it could even be made a sealed namable trait. Thus no one could apply it to new types, but it could still be used by async to match everything that should have it implemented.

external re-implmenetations of RunQueryDsl would be unaffected, for both diesel and diesel_async, as they would still need to implement RunQueryDsl for any new types.

>  It's still a breaking change in diesel-async,

Can you expand on this? The only thing I could possibly see it being a breaking change would be if someone introduced a new Query type and relied on the blanket `for T` to have it implement RunQueryDsl. That is already a deviation from diesel proper, where the external would need to explicitly implement it for their new query type, and this would technically bring it more inline with how stock diesel works.


---

Also I do see that this is ultimately a rustc 'bug', but thats how it currently works, and I don't see any signs of it getting fixed any time soon.

weiznich:
I spend a bit more time thinking about this. The variant B might work, if it doesn't regress error messages.

> I guess the marker would not need to replace the existing, but could be done in-addition to. That way nothing really changes in the public API, other than there being an additional trait that can be used to match on for diesel_async, it could even be made a sealed namable trait. Thus no one could apply it to new types, but it could still be used by async to match everything that should have it implemented.

If we do this change I personally would prefer to use it in diesel internally and also expose this trait publicly + document that this is now the preferred way to get a `RunQueryDsl` impl as this gets diesel and diesel-async closer together. It's explicitly wanted that people can apply it to their own query types. 



> Can you expand on this? The only thing I could possibly see it being a breaking change would be if someone introduced a new Query type and relied on the blanket for T to have it implement RunQueryDsl. That is already a deviation from diesel proper, where the external would need to explicitly implement it for their new query type, and this would technically bring it more inline with how stock diesel works.

Yes that's exactly what would break. Given that diesel-async is not at the 1.0 state yet that's acceptable, it's just something to note.

weiznich (approved):
I did push an update to bring this up to date with what got merged to diesel. This now is ready to be merged.

PR #228, opened 2025-04-20
Upgrade `mysql_common` and `mysql_async`

paolobarbolini:
This dependency bump includes a big reduction in number of indirect dependencies [^1] [^2] when the `mysql` feature is enabled.

[^1]: https://github.com/blackbeam/rust_mysql_common/pulls?q=is%3Apr+author%3Apaolobarbolini+
[^2]: https://github.com/blackbeam/mysql_async/pulls?q=is%3Apr+author%3Apaolobarbolini+

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,10 +23,10 @@ futures-util = { version = "0.3.17", default-features = false, features = [
 ] }
 tokio-postgres = { version = "0.7.10", optional = true }
 tokio = { version = "1.26", optional = true }
-mysql_async = { version = "0.35", optional = true, default-features = false, features = [
+mysql_async = { version = "0.36.0", optional = true, default-features = false, features = [
   "minimal-rust",
 ] }
-mysql_common = { version = "0.34", optional = true, default-features = false }
+mysql_common = { version = "0.35.3", optional = true, default-features = false }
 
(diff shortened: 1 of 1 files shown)


weiznich:
Thanks for filling :heart:

The CI seems to be fine, so lets wait on the release of mysql_async and then this seems to be ready for merging.

paolobarbolini:
`mysql_async` just came out. This should now be ready.

I've bumped the `mysql_common` patch version to align with the minimum version required by `mysql_async`. Technically  not necessary.

weiznich (approved):
Thanks for the update. Lets merge this now

> EDIT: could you also release a new version of diesel_async after merging this? I'd really appreciate getting this huge dependency cleanup out.

That's unfortunately not easily possible as the main branch of diesel-async depends on the master branch of diesel, so we need to wait for a release of the later before we could release the former without doing backports. For a the next non-patch diesel release some features are still missing, so if you want a fast release of diesel-async we likely need to backport this change to the released diesel-async version. I personally won't have the capacity to deal with that in the next few weeks due to other responsibilities.

(Also technically that's not even a minor diesel-async release but I major release due to the bumped supported minimal versions of mysql_common and mysql_async)

PR #232, opened 2025-05-08
Add support for `wasm32-unknown-unknown` target

Jaltaire:
`tokio-postgres` includes support for targeting `wasm32-unknown-unknown`, and it would be wonderful to have ORM functionality with Diesel in WASM environments.

This PR enables `diesel-async` compilation with the `postgres` feature for the `wasm32-unknown-unknown` target. Please note that I have not explored what is required to support `wasm32-unknown-unknown` for other database or connection types.

I am currently using this within a Cloudflare Workers deployment. (As an aside to anyone exploring, I have not yet been able to configure `diesel-async` to work with Hyperdrive, presumably due to some Postgres feature set that `diesel-async` leverages and that Hyperdrive is not currently supporting. Using `diesel-async` with a direct connection string without Hyperdrive, however, works as expected.)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -149,6 +149,7 @@ pub trait AsyncConnection: SimpleAsyncConnection + Sized + Send {
     /// The argument to this method and the method's behavior varies by backend.
     /// See the documentation for that backend's connection class
     /// for details about what it accepts and how it behaves.
+    #[cfg(not(target_arch = "wasm32"))]
     fn establish(database_url: &str) -> impl Future<Output = ConnectionResult<Self>> + Send;
 
     /// Executes the given function inside of a database transaction
--- a/src/pg/mod.rs
+++ b/src/pg/mod.rs
@@ -29,6 +29,7 @@ use futures_util::TryFutureExt;
 use futures_util::{Future, FutureExt, StreamExt};
 use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
+use tokio::io::{AsyncRead, AsyncWrite};
 use tokio::sync::broadcast;
 use tokio::sync::oneshot;
 use tokio::sync::Mutex;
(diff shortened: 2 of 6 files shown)


src/lib.rs
```diff
@@ -149,6 +149,7 @@ pub trait AsyncConnection: SimpleAsyncConnection + Sized + Send {
     /// The argument to this method and the method's behavior varies by backend.
     /// See the documentation for that backend's connection class
     /// for details about what it accepts and how it behaves.
+    #[cfg(not(target_arch = "wasm32"))]
     fn establish(database_url: &str) -> impl Future<Output = ConnectionResult<Self>> + Send;
```
weiznich, line 153: I would like to keep this method in all configuration. 
It should be possible to provide an implementation for this for `wasm32` as you somehow need to establish a connection there as well, right?

src/pg/mod.rs
```diff
@@ -898,6 +903,16 @@ where
...
+    #[cfg(target_arch = "wasm32")]
+    wasm_bindgen_futures::spawn_local(async move {
+        match futures_util::future::select(shutdown_rx, conn).await {
+            Either::Left(_) | Either::Right((Ok(_), _)) => {}
+            Either::Right((Err(e), _)) => {
+                let _ = error_tx.send(Arc::new(e));
+            }
+        }
+    });
+
```
weiznich, line 915: Given that the only difference here to the `tokio::spawn` call above is the name of the spawn function we maybe could refactor this to something like that to minimize code duplication:

```rust
let background_future = async move {
        match futures_util::future::select(shutdown_rx, conn).await {
            Either::Left(_) | Either::Right((Ok(_), _)) => {}
            Either::Right((Err(e), _)) => {
                let _ = error_tx.send(Arc::new(e));
            }
        }
};
#[cfg(not(target_arch = "wasm32"))]
tokio::spawn(background_future);
#[cfg(target_arch = "wasm32")]
wasm_bindgen_futures::spawn_local(background_future);
```

weiznich (commented):
Thanks for opening this PR. Generally speaking I'm open to support this.

That written, there are a few things that I would like to clarify/improve before merging.

* The inline comments
* Changing the cfgs to rather use something like `#[cfg(target_family = "wasm")]` instead to also support other wasm targets
* Adding some test/example
* Adding some sort of CI check that this works. That should at least check if it compiles, ideally run all tests. Maybe https://github.com/diesel-rs/diesel/pull/4411 could serve as an example for the testing?

LucaCappelletti94:
I looked into implementing `establish` for `wasm32` and I do not think it can be done for postgres.

Its `connect` function, the `Socket` type and the socket setup module are all gated behind the `runtime` feature, which pulls in `tokio/net`, and `tokio::net::TcpStream` does not exist on `wasm32-unknown-unknown`. I am not sure whether this limit of tokio is a real blocking one or just something they did not do yet.

There is no URL a wasm build could turn into a connection on its own at this time, I believe.

PR #238, opened 2025-06-01
Reduce futures-util

paolobarbolini:
This PR, together with https://github.com/sfackler/rust-postgres/pull/1243, reduces diesel-async's reliance on `futures-util` to only the core set of APIs from futures-rs. This speeds up compilation and also makes it clearer which imports are from futures-util and which are from futures-core.


weiznich (approved):
Thanks for submitting. Looks good for me :heart:

paolobarbolini:
This should now go into effect, as I've made a new release for all crates in the rust-postgres org: https://github.com/rust-postgres/rust-postgres/releases

PR #242, opened 2025-06-26
Fix a bug in how we handle serialization errors

### LINKED ISSUE #241: Unexpected 'AlreadyInTransaction' after serializaiton error on commit
### Versions

- **Rust:** 1.87
- **Diesel:** 2.2.11
- **Diesel_async:** 0.5.2
- **Database:** postgresql 17
- **Operating System:** linux / fedora

### Feature Flags

- **diesel:** none
- **diesel_async:** postgresql

## Problem Description
After some specific serialization errors, the connection seems to have an open transaction when I'd expect it to be clean and ready to re-use. Most serialization errors have this behavior, incurring in a rollback so you can retry the transaction but sometimes it doesn't.

I'm attempting to retry transactions that yield serialization failures on execution.

I expect the connection used on the transaction to always be unbroken and clean after running the transaction despite the transaction on itself having failed due to serialization.

Sometimes the connection is left with an open transaction, resulting in a AlreadyInTransaction on trying to use it again.

No, anything else.

### Steps to reproduce

https://github.com/lsunsi/diesel-serialization-error-corruption-reproduction

## Checklist

weiznich:
This commit fixes a bug in how we handle serialization errors with the postgres backend and transactions. It turns out that we never update the transaction manager state for `batch_execute` calls, which in turn are used for executing the transaction SQL itself. That could lead to situations in which we don't roll back the transaction, but in which we should have done that.

Fixes #241

### DIFF
--- a/examples/postgres/pooled-with-rustls/src/main.rs
+++ b/examples/postgres/pooled-with-rustls/src/main.rs
@@ -41,7 +41,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     Ok(())
 }
 
-fn establish_connection(config: &str) -> BoxFuture<ConnectionResult<AsyncPgConnection>> {
+fn establish_connection(config: &str) -> BoxFuture<'_, ConnectionResult<AsyncPgConnection>> {
     let fut = async {
         // We first set up the way we want rustls to work.
         let rustls_config = ClientConfig::with_platform_verifier();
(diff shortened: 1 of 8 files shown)


lsunsi:
> It turns out that we never update the transaction manager state for batch_execute calls, which in turn are used for executing the transaction SQL itself. 

This explains a lot, thank you @weiznich !

guissalustiano (approved):

PR #247, opened 2025-07-04
Add AsyncConnectionWrapper::into_inner

### LINKED ISSUE #213: Allow AsyncConnectionWrapper to be unwrapped
`AsyncConnectionWrapper::from` takes the ownership of `AsyncConnection`, but there isn't a way to unwrap and take the ownership back.

This can be useful when the connection wrapper is only used for a short time. For example, when used with `diesel_migrations`, if we can't unwrap the connection back, we will have to open a new connection for migrating, and the connection cannot be re-pooled after using.

Another use case is that, when running unit tests with SQLite in memory database, it is important not to re-open connection as all data will be lost.

imintz:
Fixes #213.

Adds an method which enables reuse of a wrapped async connection.

### DIFF
--- a/src/async_connection_wrapper.rs
+++ b/src/async_connection_wrapper.rs
@@ -123,6 +123,17 @@ mod implementation {
         }
     }
 
+    impl<C, B> AsyncConnectionWrapper<C, B>
+    where
+        C: crate::AsyncConnection,
+    {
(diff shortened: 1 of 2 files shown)


xtexx (approved):
Thank you!

PR #248, opened 2025-07-05
Improve pipelining for Postgres

stormshield-kg:
This MR implements suggestions from https://github.com/weiznich/diesel_async/discussions/190 to improve pipelining for Postgres.

The first commit moves the methods used by the RunQueryDsl trait in the `AsyncConnectionCore` trait.

The second commit implements `AsyncConnectionCore` for `&AsyncPgConnection`, allowing to use composed futures with pipelining by using an immutable reference to the connection to execute the concurrent queries.

### DIFF
--- a/src/pg/mod.rs
+++ b/src/pg/mod.rs
@@ -114,6 +114,48 @@ const FAKE_OID: u32 = 0;
 /// # }
 /// ```
 ///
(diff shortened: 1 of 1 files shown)


lsunsi:
@stormshield-kg I'm late to the party but wanted to say that I think this is gonna be useful to me as well. Thank you!

PR #251, opened 2025-07-13
Implement notification_stream for AsyncPgConnection

lsunsi:
### Goal
This PR aims to implement a way for an AsyncPgConnection user to receive notifications from postgresql.

### Implementation
It leverages the already present [functionality](https://docs.rs/tokio-postgres/latest/tokio_postgres/struct.Connection.html#method.poll_message) of the underlying [connection](https://docs.rs/tokio-postgres/latest/tokio_postgres/struct.Connection.html). So this PR merely wants to expose the messages.

### Limitations
As @weiznich mentioned on the [discussion](https://github.com/weiznich/diesel_async/discussions/231#discussioncomment-13142257), this will only work if the connection is present (and it might not be, depending on other constructors).

The easier fix is obviously not supporting the alternative constructors, but since we are splitting functionality it's always a good place to question if they are needed.

### This PR is a DRAFT aimed to ask questions to mantainers

### DIFF
--- a/src/pg/mod.rs
+++ b/src/pg/mod.rs
@@ -170,7 +170,7 @@ pub struct AsyncPgConnection {
     transaction_state: Arc<Mutex<AnsiTransactionManager>>,
     metadata_cache: Arc<Mutex<PgMetadataCache>>,
     connection_future: Option<broadcast::Receiver<Arc<tokio_postgres::Error>>>,
-    notification_rx: Option<mpsc::UnboundedReceiver<diesel::pg::PgNotification>>,
+    notification_rx: Option<mpsc::UnboundedReceiver<QueryResult<diesel::pg::PgNotification>>>,
     shutdown_channel: Option<oneshot::Sender<()>>,
     // a sync mutex is fine here as we only hold it for a really short time
     instrumentation: Arc<std::sync::Mutex<DynInstrumentation>>,
@@ -509,7 +509,7 @@ impl AsyncPgConnection {
     async fn setup(
         conn: tokio_postgres::Client,
         connection_future: Option<broadcast::Receiver<Arc<tokio_postgres::Error>>>,
-        notification_rx: Option<mpsc::UnboundedReceiver<diesel::pg::PgNotification>>,
+        notification_rx: Option<mpsc::UnboundedReceiver<QueryResult<diesel::pg::PgNotification>>>,
         shutdown_channel: Option<oneshot::Sender<()>>,
         instrumentation: Arc<std::sync::Mutex<DynInstrumentation>>,
     ) -> ConnectionResult<Self> {
@@ -731,7 +731,7 @@ impl AsyncPgConnection {
 
     pub fn notification_stream(
         &mut self,
-    ) -> impl futures_core::Stream<Item = diesel::pg::PgNotification> + '_ {
+    ) -> impl futures_core::Stream<Item = QueryResult<diesel::pg::PgNotification>> + '_ {
         match &mut self.notification_rx {
             None => Either::Left(futures_util::stream::pending()),
             Some(rx) => Either::Right(futures_util::stream::unfold(rx, async |rx| {
@@ -987,7 +987,7 @@ fn drive_connection<S>(
     mut conn: tokio_postgres::Connection<tokio_postgres::Socket, S>,
 ) -> (
     broadcast::Receiver<Arc<tokio_postgres::Error>>,
-    mpsc::UnboundedReceiver<diesel::pg::PgNotification>,
+    mpsc::UnboundedReceiver<QueryResult<diesel::pg::PgNotification>>,
     oneshot::Sender<()>,
 )
 where
@@ -996,23 +996,25 @@ where
     let (error_tx, error_rx) = tokio::sync::broadcast::channel(1);
     let (notification_tx, notification_rx) = tokio::sync::mpsc::unbounded_channel();
     let (shutdown_tx, mut shutdown_rx) = tokio::sync::oneshot::channel();
+    let mut conn = futures_util::stream::poll_fn(move |cx| conn.poll_message(cx));
 
     tokio::spawn(async move {
-        let mut conn = futures_util::stream::poll_fn(|cx| conn.poll_message(cx));
-
         loop {
             match futures_util::future::select(&mut shutdown_rx, conn.next()).await {
                 Either::Left(_) | Either::Right((None, _)) => break,
                 Either::Right((Some(Ok(tokio_postgres::AsyncMessage::Notification(notif))), _)) => {
-                    let _: Result<_, _> = notification_tx.send(diesel::pg::PgNotification {
+                    let _: Result<_, _> = notification_tx.send(Ok(diesel::pg::PgNotification {
                         process_id: notif.process_id(),
                         channel: notif.channel().to_owned(),
                         payload: notif.payload().to_owned(),
-                    });
+                    }));
                 }
                 Either::Right((Some(Ok(_)), _)) => {}
                 Either::Right((Some(Err(e)), _)) => {
-                    let _ = error_tx.send(Arc::new(e));
+                    let e = Arc::new(e);
+                    let _: Result<_, _> = error_tx.send(e.clone());
+                    let _: Result<_, _> =
+                        notification_tx.send(Err(error_helper::from_tokio_postgres_error(e)));
                     break;
                 }
             }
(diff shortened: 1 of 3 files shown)


src/pg/mod.rs
```diff
@@ -969,27 +990,42 @@ async fn drive_future<R>(
...
-            Either::Left(_) | Either::Right((Ok(_), _)) => {}
-            Either::Right((Err(e), _)) => {
-                let _ = error_tx.send(Arc::new(e));
+        let mut conn = futures_util::stream::poll_fn(|cx| conn.poll_message(cx));
+
+        loop {
+            match futures_util::future::select(&mut shutdown_rx, conn.next()).await {
+                Either::Left(_) | Either::Right((None, _)) => break,
+                Either::Right((Some(Ok(tokio_postgres::AsyncMessage::Notification(notif))), _)) => {
+                    let _ = notification_tx.send(diesel::pg::PgNotification {
```
lsunsi, line 1013: How should this error be handled? Technically not being able to send here just means the consumer is dead, which I think might happen in a race condition with all the clients being dropped while we wait for the messages.

src/pg/mod.rs
```diff
@@ -969,27 +990,42 @@ async fn drive_future<R>(
...
     broadcast::Receiver<Arc<tokio_postgres::Error>>,
+    broadcast::Receiver<diesel::pg::PgNotification>,
     oneshot::Sender<()>,
 )
 where
     S: tokio_postgres::tls::TlsStream + Unpin + Send + 'static,
 {
     let (error_tx, error_rx) = tokio::sync::broadcast::channel(1);
-    let (shutdown_tx, shutdown_rx) = tokio::sync::oneshot::channel();
+    let (notification_tx, notification_rx) = tokio::sync::broadcast::channel(1);
```
lsunsi, line 1003: I added the broadcast channel because it felt easier due to it being multi-receiver, but I'm not certain it's the right call because it means we could loose messages. But pondering the right approach here raises a lot of questions in my head.

1. We don't want to backpressure or hold all messages in case the user does not consume them. This means the proper treatment would be to just drop all messages not being consumed. Does this make sense?
2. What should be the behavior of calling `notification_stream` twice? Should it be `&mut` so there can only be one iterator at a time, or should we allow multiple duplicate streams? Upon writing this, I think maybe the first approach is better and more inline with sync diesel impl. Any thoughts?

src/pg/mod.rs
```diff
@@ -969,27 +990,42 @@ async fn drive_future<R>(
...
+    let (shutdown_tx, mut shutdown_rx) = tokio::sync::oneshot::channel();
 
     tokio::spawn(async move {
-        match futures_util::future::select(shutdown_rx, conn).await {
-            Either::Left(_) | Either::Right((Ok(_), _)) => {}
-            Either::Right((Err(e), _)) => {
-                let _ = error_tx.send(Arc::new(e));
+        let mut conn = futures_util::stream::poll_fn(|cx| conn.poll_message(cx));
```
lsunsi, line 1007: Here I could use stream::poll_fn to build the stream or just poll_message with std::future::poll_fn.

I went for the former because it felt more appropriate but I could replace it with the latter if you feel using std is better (or some other reason). The stream is not really needed here it seems.

src/pg/mod.rs
```diff
@@ -724,6 +730,21 @@ impl AsyncPgConnection {
             .unwrap_or_else(|p| p.into_inner())
             .on_connection_event(event);
     }
+
+    pub fn notification_stream(
```
lsunsi, line 734: So you mentioned the behavior for not having a Connection should be having a `pending stream`. The way I did makes it so the stream is just `empty`. I think the empty (besides being easier to code) makes more sense because we could never receive a notification and will point the user to the documentation explaining why it's empty. What do you think?

src/pg/mod.rs
```diff
@@ -477,6 +479,7 @@ impl AsyncPgConnection {
             conn,
             None,
             None,
+            None,
```
lsunsi, line 482: This is the function we could think about deprecating right? It does look weird having all these options. Can you share the rationale behind this construction addition to the project?

src/pg/mod.rs
lsunsi: Actually, on second thought, the conn only gets notifications it has subscribed (listened) to. So maybe my worry is unwarranted and the correct implementation would be mpsc instead of broadcast.

lsunsi: I think since the user asked for the `listen`, the behavior makes a lot of sense. It's not something that can blow up on it's own, it really requires misuse to behave badly.

src/pg/mod.rs
dullbananas: I need the ability to use both `tokio_postgres::connect` and notifications, so there should be a constructor that takes both `Client` and `Connection`.

weiznich: @dullbananas That [method](https://docs.rs/diesel-async/latest/diesel_async/struct.AsyncPgConnection.html#method.try_from_client_and_connection) already exists, so I'm not sure what you want to add here.

src/pg/mod.rs
lsunsi: Makes sense. Thanks

src/pg/mod.rs
lsunsi: Got it. The only way I could achieve this behavior was through a manual stream implementation, do you know of a different way? Example: 910738116c30835d4571ea0170d25342c3598bc5

src/pg/mod.rs
```diff
@@ -724,6 +730,21 @@ impl AsyncPgConnection {
...
+    pub fn notification_stream(
+        &self,
+    ) -> impl futures_core::Stream<Item = QueryResult<diesel::pg::PgNotification>> {
+        futures_util::stream::unfold(
+            self.notification_rx.as_ref().map(|rx| rx.resubscribe()),
+            |rx| async {
+                let mut rx = rx?;
+                match rx.recv().await {
+                    Ok(notification) => Some((Ok(notification), Some(rx))),
+                    Err(_) => todo!(),
```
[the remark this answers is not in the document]
lsunsi: Got it. Turns out it was unneeded 910738116c30835d4571ea0170d25342c3598bc5

src/pg/mod.rs
```diff
@@ -724,6 +730,21 @@ impl AsyncPgConnection {
             .unwrap_or_else(|p| p.into_inner())
             .on_connection_event(event);
     }
+
+    pub fn notification_stream(
+        &self,
```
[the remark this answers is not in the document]
lsunsi: Perfect, done! 910738116c30835d4571ea0170d25342c3598bc5

lsunsi:
@weiznich I updated the impl in the right direction, I only have one more major question I'll add as a comment. Tests and doc still missing

src/pg/mod.rs
```diff
@@ -724,6 +728,28 @@ impl AsyncPgConnection {
             .unwrap_or_else(|p| p.into_inner())
             .on_connection_event(event);
     }
+
+    pub fn notification_stream(
+        &mut self,
+    ) -> impl futures_core::Stream<Item = diesel::pg::PgNotification> + '_ {
```
lsunsi, line 734: It's about this signature. I changed it to Stream<Notification> to match the channel, but now I think maybe it's wrong because you said it should be Stream<QueryResult<Notification>>.

That said, the poll_messages function exposes Ok(Notification | Notice) | Err, so if I get an error I can't be sure it's from a notification, a notice, or anything. It's basically just an error.

So it feels to me that I should either output the 3 possible cases, or just pick the one I want (notification). But I'm not sure of this. Anyway, I think diesel sync might have gone through the same question?

weiznich: I think it needs to be `Stream<QueryResult>` as we currently miss one important point: We still need to poll the `drive_connection` future as well as part of the stream, as otherwise we won't get any new notifications while the user polls on the channel we returned. I think that connection future might return an error (via the error channel), that then must be returned from the notification stream.

src/pg/mod.rs
weiznich: Maybe just return a `futures::either::Either` stream instead? https://docs.rs/futures/0.3.31/futures/future/enum.Either.html. We could use the channel as one variant and a `futures::stream::pending` as the other variant (https://docs.rs/futures/0.3.31/futures/prelude/stream/fn.pending.html)

(although I'm not 100% sure if that's all in the futures_util crate or not)

lsunsi:
@weiznich I think current implementation might be good enough. If we're getting closer, I'll work on the test next

weiznich (approved):
Looks good for me, thanks for working on this :+1:

Issue #252, opened 2025-07-16
Serialization failure on commit issues a rollback without open transaction


lsunsi:
## Setup

So reproduce this problem you need to cause a serialization failure on commit. On postgres it'll issue a notice telling you that you tried to rollback but no transaction is open.

### Versions

- **Rust:** 1.88
- **Diesel:** 2.2.0
- **Diesel_async:** 0.6.1
- **Database:** Postgres
- **Operating System:** Linux

### Feature Flags

- **diesel:**
- **diesel_async:** postgres

## Problem Description

Diesel is issuing  rollback statement after the transaction is already closed (due to serialization error on commit).
Note that the rollback is indeed necessary if the serialization error came from any other statement other than commit.

### What are you trying to accomplish?

I'm trying to use diesel async in a highly async context with lots of concurrent transactions that can fail on serialization isolation level.

### What is the expected output?

No rollback is ever issue without an open transaction.

### What is the actual output?

Rollback is issue without open transaction because the failed commit already closes the transaction.

### Are you seeing any additional errors?

I'm seeing a rust-postgres log notice coming from the database that's telling me I'm doing something wrong.
log: `WARNING: there is no transaction in progress`

### Steps to reproduce

I added a test that intends to show what I'm seeing.
https://github.com/lsunsi/diesel_async/blob/reproduce-rollback-after-commit/tests/transactions.rs#L110

## Checklist

- [x] I have already looked over the [issue tracker](https://github.com/weiznich/diesel_async/issues) for similar possible closed issues.
 
- [x] This issue can be reproduced on Rust's stable channel. (Your issue will be
  closed if this is not the case)
- [x] This issue can be reproduced without requiring a third party crate

### What I think generated this issue
Coming from this issue https://github.com/weiznich/diesel_async/issues/241 the PR that fixed it also added this subtle bug. The reason I think this happened is because the serialization error on commit is different from other serialization errors.

In case this happens on UPDATE for example, the behavior should be to reset the diesel transaction depth while issuing a rollback. In the case of it happening on a COMMIT, the transaction depth should be considered as if the commit had happened (should be reset) but the rollback should not be issued in this case.

cc @lochetti

weiznich:
Thanks for creating this issue and providing a reproduction. Unfortunately I do not have the capacity yet to look into this in detail, so this needs to wait until I have more time. If someone wants to look at this before that, I'm happy to accept a PR with a fix.

PR #253, opened 2025-07-19
Fix rollback being attempted on no transaction because commit already rolled it back

lsunsi:
### Problem
The issue is better described in the linked issue.
Fixes https://github.com/weiznich/diesel_async/issues/252.

### Solution
I think the commit succeeding and failing specifically for SerializationError means the same for the transaction manager: that it should adjust it's internal state and nothing else.

### How I found out about the issue
I'm running diesel_async on a highly concurrent web-server. On 0.5.x I saw that some transactions were left open on the transaction manager (in memory state) so that the next transaction starting would cause a failure (AlreadyInTransaction).

0.6.x fixed that guaranteeing that serialization failures on commit would generate a rollback. No AlreadyInTransaction was ever seen again, but then we started getting "there is no transaction in progress" from PostgreSQL.

My reasoning is that on 0.5.x, the connection behavior was correct but the in-memory state was not.
Then in 0.6.x, the connection behavior is correct, the in-memory was also correct, but an extra ROLLBACK command was issued, which is incorrect behavior.

Then I think this PR leaves things in a correct state, as the test shows.

### Shortcomings
I'm not sure about the effect of this PR on nested transaction, especially since I don't use them and don't know much about them. Maybe someone can shed some light if this change is enough or more work needs to be done to avoid breakage in these scenarios.

cc @lochetti

### DIFF
--- a/tests/transactions.rs
+++ b/tests/transactions.rs
@@ -191,7 +191,7 @@ async fn commit_with_serialization_failure_already_ends_transaction() {
     };
 
     let (res, res1) = tokio::join!(res, res1);
-    let _ = diesel::sql_query("DROP TABLE users3")
+    let _ = diesel::sql_query("DROP TABLE users4")
         .execute(&mut conn1)
         .await;
 


src/transaction_manager.rs
```diff
@@ -358,7 +361,7 @@ where
         let is_broken = conn.transaction_state().is_broken.clone();
 
         match Self::critical_transaction_block(&is_broken, conn.batch_execute(&commit_sql)).await {
-            Ok(()) => {
+            a @ (Ok(()) | Err(Error::DatabaseError(SerializationFailure, _))) => {
```
weiznich, line 364: I'm not sure if that logic applies to other backends as well. Given that AnsiTransaction manager is also used for the Mysql backend I would rather prefer a solution that updates [this](https://github.com/weiznich/diesel_async/blob/main/src/pg/mod.rs#L383-L395) function to only set `requires_rollback_maybe_up_to_top_level` for cases where we do not commit? That could be done by having a flag in the transaction manager indicating that we actually execute commit now or something like that. So possibly:

* Add a bool `is_commit` flag to `AnsiTransactionManager`
* Set that flag to true before the `batch_execute` in the line above, set it to false in the end of `critical_transaction_block`
* Change `update_transaction_manager_status` to check for that flag set `requires_rollback_myabe_up_to_top_level` for it. If that flag is set to false, we won't perform a rollback below.

For reference: The [diesel](https://github.com/diesel-rs/diesel/blob/master/diesel/src/pg/connection/mod.rs#L282-L334) version defers to libpq for that information, but for diesel-async we need to track that on our own.

At some point we should at least try to unify as much of the transaction manager logic as possible, but that's something for future improvements.

lsunsi: @weiznich Did you have something like this in mind? https://github.com/weiznich/diesel_async/pull/253/commits/a175eb264f3e98e8548a9c4952b316a5a8808744

Having the boolean used to tell we're in a commit and avoiding setting the other boolean flag makes a lot of sense to me, but the fact that we still have to change the `commit_transaction` function in order to make sure the transaction state is correct (we need to decrement it without rolling back) felt a bit off to me.

I don't know maybe it's fine, you'll be the better judge.
What do you think?

tests/transactions.rs
```diff
@@ -104,3 +104,125 @@ async fn concurrent_serializable_transactions_behave_correctly() {
...
+                    Ok::<_, diesel::result::Error>(())
+                })
+            })
+            .await;
+        barrier_2_for_tx2.wait().await;
+        res
+    };
+
+    let (res, res1) = tokio::join!(res, res1);
+    let _ = diesel::sql_query("DROP TABLE users3")
```
weiznich, line 194: That needs to be `users4` right?


```suggestion
    let _ = diesel::sql_query("DROP TABLE users4")
```

weiznich (approved):
Looks good to me beside that minor correction for the test

Issue #249, opened 2025-07-07
Compilation error when using pipelining


stormshield-kg:
### Versions

- **Rust:**: 1.88.0
- **Diesel:**: commit `cdf7c51c35da7ebc6fb12e71857bfba8487c20cd`
- **Diesel_async:**: commit 0e703ba9446213f91acd4a2371a825421113c321
- **Operating System**: Ubuntu 22

## Problem Description

The following code doesn't compile:

```rust
fn f(conn: &mut crate::AsyncPgConnection) {
    use diesel::sql_types::Integer;
    use diesel::IntoSql;
    use futures_util::future::try_join;
    use futures_util::FutureExt;

    use crate::RunQueryDsl;

    let f1 = diesel::select(1_i32.into_sql::<Integer>()).get_result::<i32>(conn);
    let f2 = diesel::select(2_i32.into_sql::<Integer>()).get_result::<i32>(conn);

    // This doesn't work
    let _ = async { try_join(f1, f2).await }.boxed();
}
```

<details> 

<summary>
Error
</summary>

```
error[E0308]: mismatched types
    --> src/pg/mod.rs:1145:17
     |
1145 |         let _ = async { try_join(f1, f2).await }.boxed();
     |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ one type is more general than the other
     |
     = note: expected struct `Map<StreamFuture<Pin<Box<Map<Pin<Box<dyn Stream<Item = ...> + Send>>, ...>>>>, ...>`
                found struct `Map<StreamFuture<Pin<Box<Map<Pin<Box<dyn Stream<Item = ...> + Send>>, ...>>>>, ...>`
note: the lifetime requirement is introduced here
    --> /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/futures-util-0.3.31/src/future/future/mod.rs:516:23
     |
516  |         Self: Sized + Send + 'a,
     |                       ^^^^
     = note: the full name for the type has been written to '/home/runner/Desktop/diesel_async/target/debug/deps/diesel_async-0e1fc68988d593f9.long-type-4500313906557700460.txt'
     = note: consider using `--verbose` to print the full type name to the console

error[E0308]: mismatched types
    --> src/pg/mod.rs:1145:17
     |
1145 |         let _ = async { try_join(f1, f2).await }.boxed();
[...]

PR #262, opened 2025-09-19
Use custom `Future` combinators to avoid GAT errors

stormshield-kg:
Fixes https://github.com/weiznich/diesel_async/issues/249.

The compilation errors were caused by the lifetimes when using closures with `Future` combinators in the `RunQueryDsl` trait.
This PR fixes the GAT errors by redefining custom `Future` combinators to limit closure usage.


weiznich:
Thanks again for figuring this out :heart: 

I did some refactoring to use the future combinators from `futures-utils` instead of the hand written ones. I'm still a bit surprised that this works but using a plain type def fails for whatever reason with a impl not general enough error. Seems like another rustc limitation for me. Maybe I should open a bug about that someday…

PR #284, opened 2026-04-02
Use real async closures for transactions

weiznich:
This commit changes all transaction related functions to use real async closures instead of closures returning a boxed future. This resolves a long standing issue in expressing the right bounds for this closure and the relevant return type.

A huge thanks to lcnr who helped to figure out a working workaround for all the involed bugs in rustc: https://github.com/lcnr/random-rust-snippets/issues/26


tlowerison:
Just want to say it's awesome to see the `transaction` function finally make it all the way to not having to box futures and still get the correct lifetime bounds. Glad to see `scoped-futures` phased out and this crate continue to evolve!

PR #293, opened 2026-06-10
Map restrict and exclusion violations in `from_tokio_postgres_error` used in PG18

ajesipow:
I ran into this issue when upgrading from Postgres 17 to Postgres 18:

```
"update or delete on table \"user\" violates RESTRICT setting of foreign key constraint \"sales_order_user_fkey\" on table \"sales_order\"
```

In Postgres 17, when deleting a row with a FK constraint and an `ON DELETE RESTRICT` action a FK violation was triggered (and correctly propagated as such via `diesel::result::DatabaseErrorKind`).

This changed apparently with in v18 where a `RESTRICT VIOLATION` is thrown in such cases. This `SqlState`  wasn't mapped yet in `from_tokio_postgres_error` and therefore fell through to `Unknown`.

I also added `EXCLUSION_VIOLATION` for consistency.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -115,7 +115,9 @@ jobs:
             sudo apt-get install -y postgresql-common
             sudo /usr/share/postgresql-common/pgdg/apt.postgresql.org.sh -y
             sudo apt-get install -y "postgresql-$PG_VERSION"
-            sudo systemctl restart "postgresql@$PG_VERSION-main"
+            sudo pg_dropcluster --stop "$PG_VERSION" main 2>/dev/null || true
+            sudo pg_createcluster "$PG_VERSION" main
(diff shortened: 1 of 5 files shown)


weiznich:
Thanks for opening this PR. Would it be possible to add a test case that triggers the new cases?

ajesipow:
I added tests that cover all of the concrete errors from `from_tokio_postgres_error`.

The `RESTRICT VIOLATION` test checks the postgres version to cover the change in behaviour since version 18 (added two checks to CI for PG17 and PG18 on Ubuntu to excercise these two paths in the test).

I also removed the `echo "host    all             all             127.0.0.1/32            md5" > sudo tee -a /etc/postgresql/10/main/pg_hba.conf` line in `ci.yml` as it seemed unnecessary.

weiznich (approved):
Looks good now, thanks for fixing this. 

I just pushed a few CI fixes on top to let the CI pass without problems

PR #294, opened 2026-06-18
Use get_result instead of execute for pool connection ping

### LINKED ISSUE #139: The conneciton pool returns corrupt connections
### Versions

- **Rust: rustc 1.75.0 (82e1608df 2023-12-21)**
- **Diesel: 2.1.0**
- **Diesel_async: 0.4.1**
- **Database: mysql 5.7.26-log**
- **Operating System： linux**

### Feature Flags

- **diesel: ["mysql_backend", "time"]**
- **diesel_async: ["mysql", "deadpool"]**

## Problem Description

After running for a few days, some connections in the connection pool will return an error

Query records in the database

No error

Got an error: "Unexpected end of row"

Input/output error: can't parse: buf doesn't have enough data

### Steps to reproduce

It's hard to reproduce the problem, after running for a few days, some connections in the pool will return errors. I think those connections are already corrupt but still can pass the health check of the connection pool. So I did a contrast experiment, when I found the connection got corrupt, the `ping` responded with ok, but the `ping2` responded with an error "buf doesn't have enough data"

```rust

use diesel_async::{pooled_connection::deadpool::Pool, AsyncMysqlConnection};

type MysqlPool = Pool<AsyncMysqlConnection>;

pub struct AppState {
    pub mysql_pool: MysqlPool,
}


async fn ping(State(state): State<Arc<AppState>>) -> Result<impl IntoResponse> {
    let mut conn = state.mysql_pool.get().await?;
    let x: i32 = diesel::select(1_i32.into_sql::<diesel::sql_types::Integer>())
        .first(&mut conn)
        .await?;

    Ok(Json(x))
}

async fn ping2(State(state): State<Arc<AppState>>) -> Result<impl IntoResponse> {
    let mut conn = state.mysql_pool.get().await?;
    let x = diesel::select(1_i32.into_sql::<diesel::sql_types::Integer>())
        .execute(&mut conn)
        .await?;

    Ok(Json(x.rows_affected))
}

```

## Checklist

- [ ] I have already looked over the [issue tracker](https://github.com/weiznich/diesel_async/issues) for similar possible closed issues.
 
- [ ] This issue can be reproduced on Rust's stable channel. (Your issue will be
- [ ] This issue can be reproduced without requiring a third party crate

DhineshKrishnan1206:
## Summary

- Changes the `Verified` recycling method's ping from `.execute()` to `.get_result::<i32>()` so the response row is actually read back from the server
- This detects connections with corrupt internal read buffers that silently pass `.execute()` (which only reads the row count) but fail on real queries with "Unexpected end of row" / "buf doesn't have enough data" errors
- Updates the corresponding trait bounds from `ExecuteDsl` to `LoadQuery` across all pool backends (bb8, deadpool, mobc) and the r2d2 connection wrapper

Fixes #139

## Context

As discussed in #139, the current ping uses `.execute("SELECT 1")` which sends the query and reads the affected row count, but never reads actual row data. A connection whose read buffer contains leftover bytes from a previous partially-consumed response will pass this check, get returned to the pool, and then fail on the next real query.

Using `.get_result::<i32>()` forces the connection to deserialize a row, which will surface buffer corruption immediately during the health check rather than on a subsequent application query.

## Test plan

- [x] `cargo check --features "postgres bb8 deadpool mobc r2d2"` — compiles cleanly
- [x] `cargo check --features "mysql bb8 deadpool mobc"` — compiles cleanly
- [x] `cargo check --features "sqlite bb8 deadpool mobc r2d2"` — compiles cleanly

### DIFF
--- a/src/async_connection_wrapper.rs
+++ b/src/async_connection_wrapper.rs
@@ -315,8 +315,8 @@ pub(crate) mod implementation {
         C: crate::AsyncConnection<Backend = <Self as diesel::Connection>::Backend>
             + crate::pooled_connection::PoolableConnection
             + 'static,
-        diesel::dsl::select<diesel::dsl::AsExprOf<i32, diesel::sql_types::Integer>>:
-            crate::methods::ExecuteDsl<C>,
+        for<'a> diesel::dsl::select<diesel::dsl::AsExprOf<i32, diesel::sql_types::Integer>>:
+            crate::methods::LoadQuery<'a, C, i32>,
         diesel::query_builder::SqlQuery: crate::methods::ExecuteDsl<C>,
     {
         fn ping(&mut self) -> diesel::QueryResult<()> {
--- a/src/pooled_connection/bb8.rs
+++ b/src/pooled_connection/bb8.rs
@@ -67,8 +67,8 @@ pub type RunError = bb8::RunError<super::PoolError>;
 impl<C> ManageConnection for AsyncDieselConnectionManager<C>
 where
     C: PoolableConnection + 'static,
-    diesel::dsl::select<diesel::dsl::AsExprOf<i32, diesel::sql_types::Integer>>:
-        crate::methods::ExecuteDsl<C>,
+    for<'a> diesel::dsl::select<diesel::dsl::AsExprOf<i32, diesel::sql_types::Integer>>:
+        crate::methods::LoadQuery<'a, C, i32>,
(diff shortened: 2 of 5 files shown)


weiznich (approved):
Looks good, thanks for submitting :heart:

This can be merged after a rebase on top of #293 as that one contains the fix for the failing CI job. I will take care of this later today.

PR #298, opened 2026-09-07
Bump minimal supported `mysql_async` version to `0.37`

LucaCappelletti94:
The `mysql` feature currently requires `mysql_async 0.36`, whose prepared-statement cache pulls in `lru 0.16`. That version of `lru` is flagged by `RUSTSEC-2026-0253` for a potential use-after-free from missing panic safety in `LruCache::pop`, and the fix only landed in `lru 0.18.2`. No `0.36` release of `mysql_async` permits that, so any project running cargo audit against a `diesel-async` MySQL graph is stuck with the advisory.

This, therefore, bumps the requirement to `mysql_async 0.37`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 
 * Add support for `#[derive(MultiConnection)]`
 * Bump minimal supported Rust version to 1.88.0
+* Bump minimal supported `mysql_async` version to 0.37 (pulls in a patched `lru`, RUSTSEC-2026-0253)
 
 ## [0.9.2] - 2026-06-19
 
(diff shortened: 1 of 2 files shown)


weiznich (approved):
I'm still unsure if we should issue a release now or wait a few more weeks until hopefully Diesel 2.4 is finally released?

LucaCappelletti94:
> I'm still unsure if we should issue a release now or wait a few more weeks until hopefully Diesel 2.4 is finally released?

Does `diesel_async` at this time use things from the newer diesel? If so, I think waiting would be best so as to avoid picking apart what to publish and what not.

weiznich:
Yes, the main branch of diesel-async currently contains additional code for `#[derive(MultiConnection)]` that requires a 2.4 release from diesel_derives.

On the other hand it should be trivial to backport this as it is only a small change to `Cargo.toml`

